PR #5, opened 2016-10-18
Scripts

sherter:
I wasn't sure if this task was assigned to somebody. So here it is ;)

There were some problems with the jmh plugin (see https://github.com/melix/jmh-gradle-plugin/issues/81)
Since we must have a 'build' script, we can't have the default 'build' directory - and jmh doesn't handle that properly. I disabled the plugin on travis for now, until we find a solution or the issue gets fixed.


sgraf812:
Thanks!

> Since we must have a 'build' script, we can't have the default 'build' directory

That's unfortunate...

sgraf812:
>  jmh doesn't handle that properly

That might have to do with the fact that `buildDir` is only set after the `plugins{}` section. I tried to move it to the top of the file, but then gradle would complain that nothing may precede `plugins{}`. 

Maybe there is some way to set `buildDir` in a way that it's also visible for the jmh plugins?

sherter:
There is nothing we can really do about that, except fixing the plugin. If the maintainer of the plugin doesn't respond in a timely manner, I can look into that. Besides, do we really need that plugin anyway? I don't know exactly what it does, but it seems like it's running benchmarks on the JVM - and we compile down to x86-assembler?!

sherter:
Oh, we could try to apply the plugin after setting `buildDir` (with the old gradle plugin application mechanism).

sgraf812:
> and we compile down to x86-assembler?!

Ahh right x). 

> Oh, we could try to apply the plugin after setting buildDir (with the old gradle plugin application mechanism).

Did exactly that, opening a PR just now...

PR #6, opened 2016-10-18
Fix jmh

sgraf812:
Well, if it's needed at least. Otherwise we can just park it here in case we need it some time later.

(Continues the discussion of PR #5)


sgraf812:
Don't know why benchmarks aren't found, so I'll put this on hold.

Edit: Well, no I know. I forgot to `git add` the new folder.

PR #9, opened 2016-10-18
Implement and test `--echo` option

parttimenerd:


sgraf812:
That's cool and all, but I'm not so sure about the test...

Resources should be loaded at runtime from the jar instead of hardcoding a path to them in the test suite. That's probably not easily possible without further extracting even more trivial logic from `Cli` (e.g. extract a method `echo`, taking an `InputStream` and an `OutputStream`, redirecting all content), which will make the logic to test even more trivial.

Given that there's practically nothing to test at all, I'd say that the fact that paths are hardcoded just makes the test suite more brittle (e.g. moving the resources will result in an error at runtime rather than compile time) while providing only little value.

parttimenerd:
The first version of this code used a temporary file which is generated at runtime. Do you think this approach would be better?

sgraf812:
I think so. But really I don't think a test is necessary. Basically once it type-checks, it shouldn't break. Especially because we probably won't touch this code ever again.

parttimenerd:
It's just a dummy test which allows the setup of directories and discussions about later tests.
It's not intended to be there for weeks.

sgraf812:
Convinced :)

sherter:
We certainly won't put the source files we want to compile (or echo) with our compiler _into_ the compiler's jar file, so there is nothing to load from the jar here really.

That said, I don't like the test very much either. A test for the `build` and `run` scripts would be much more valuable on this level, at least to prevent regressions.

src/test/java/minijava/CliTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+	/**
+	 * Tests the `--echo` option of the cli by calling the `run` script
+     */
+	@Test
+	public void echo() throws Exception {
+		Process process = Runtime.getRuntime().exec(new String[]{"./run", "--echo", "non_existing_file"});
+		process.waitFor();
+		assertEquals("Error code for invalid --echo call", 1, process.exitValue());
+		File tmpFile = File.createTempFile("tmp", "tmp");
```
sherter, line 23: nit-pick: I prefer the java.nio.file-API when working with files ;)

sherter (changes requested):
Please remove empty_file and non_empty_file, otherwise I'm pretty happy with it now!

src/test/java/minijava/CliTest.java
```diff
@@ -0,0 +1,32 @@
...
+		FileWriter writer = new FileWriter(tmpFile);
+		writer.write("t");
+		writer.close();
+		process = Runtime.getRuntime().exec(new String[]{"./run", "--echo", tmpFile.getAbsolutePath()});
+		process.waitFor();
+		assertEquals(0, process.exitValue());
+		assertEquals('t', process.getInputStream().read());
```
sherter, line 30: Temp-files should be auto-removed at some point, but it's always good to clean up...

sherter (approved):

PR #10, opened 2016-10-19
Other dir testcase

parttimenerd:
This test case checks whether or not the `run` script can be called from a temporary directory (as required).

The last commit of this PR fixes the `run` script to allow this testcase tu succeed.


run
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/env sh
 
+SCRIPT_PATH=$(readlink -f "$0")
+PROJECT_DIR=$(dirname "$SCRIPT_PATH")
+cd $PROJECT_DIR
```
sherter: > Also weder im Skript noch im Compiler selbst das Verzeichnis wechseln

If you change the directory, the argument passed to --echo will be relative to the project dir instead of the current working directory. Instead simply call the `minijavac` bash script with its absolute path (I should have done that in the first place, my bad).

`"$PROJECT_DIR/build_dir/install/minijavac/bin/minijavac" "$@"`

sherter (commented):
Thanks for adding a test case for that scenario. This test _is_ valuable after all! :wink:
However, now we have two PRs with (partially) the same content. Please close one of them (and if not too much of a hassle, squash some of the commits that logically belong together).

src/test/java/minijava/CliTest.java
```diff
@@ -14,20 +17,33 @@
...
 		FileWriter writer = new FileWriter(tmpFile);
 		writer.write("t");
 		writer.close();
 		process = Runtime.getRuntime().exec(new String[]{"./run", "--echo", tmpFile.getAbsolutePath()});
 		process.waitFor();
 		assertEquals(0, process.exitValue());
 		assertEquals('t', process.getInputStream().read());
-		tmpFile.delete();
+
+		File tmpDirectory = Files.createTempDirectory("test").toFile();
```
sherter: It's preferable to have multiple small tests instead of one big one. The advantage is that you can immediately see in the test report what's wrong (if the test cases have a expressive names, say "runEchoWithProjectDirAsWorkingDirectory" and "runEchoWithSomeTempDirAsWorkingDirectory")

parttimenerd:
The issues, you pointed out, should be resolved.

run
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env sh
 
+SCRIPT_PATH=$(readlink -f "$0")
+PROJECT_DIR=$(dirname "$SCRIPT_PATH")
+
 # use the script generated by Gradle (does all the relative-path-magic for the dependencies)
-build_dir/install/minijavac/bin/minijavac "$@"
+$PROJECT_DIR/build_dir/install/minijavac/bin/minijavac "$@"
```
sherter: Please don't forget the quotes around the path (`"$PROJECT_DIR/build_dir/install/minijavac/bin/minijavac"`)
in case somebody uses whitespace in there directory names (shame on them!) :wink:

sherter:
Looks good to me!

run
parttimenerd: I'm writing a test for it…

PR #11, opened 2016-10-19
Print the usage info for invalid arguments

parttimenerd:


src/main/java/minijava/Cli.java
```diff
@@ -20,15 +21,34 @@
...
 
     Cli(OutputStream out, OutputStream err, String... args) {
         this.out = new PrintStream(out);
         this.err = new PrintStream(err);
-        this.jCommander = new JCommander(this, args);
+        JCommander commander;
+        try {
+            commander = new JCommander(this, args);
+        } catch (ParameterException ex){
+            commander = null;
```
sherter: This line isn't necessary, you overwrite it in line 41

src/main/java/minijava/Cli.java
```diff
@@ -20,15 +21,34 @@
     private final PrintStream out;
     private final PrintStream err;
     private final JCommander jCommander;
+    /**
+     * Did the command argument parsing fail?
+     */
+    private final boolean invalidArguments;
 
 
     Cli(OutputStream out, OutputStream err, String... args) {
```
sherter: I didn't really like it before, but now that we are adding more functionality... 

Constructors should not cause side effects. I would prefer to add a static factory method `static Cli create(OutputStream, OutputStream, String... args)` and make the constructor private. And maybe rethrow the `ParameterException` from that static factory method and handle it where the `Cli.create` method is called. Or even better, create a new Exception class, so we don't tie us to that specific argument parsing library.

parttimenerd: You're right. I'm editing my PR accordingly.

sgraf812:
While I think this is a little overengineered, it's pretty to safe to include as is.

PR #14, opened 2016-10-25
Add a lexer implementation

parttimenerd:


src/main/java/minijava/lexer/BasicLexerInput.java
```diff
@@ -0,0 +1,60 @@
+package minijava.lexer;
+
+import java.io.IOException;
+import java.io.InputStream;
+import minijava.MJError;
+
+/** A basic lexer input stream without advanced buffering. */
```
sherter, line 8: Why exactly don't you use buffering? This kills performance. At least wrap the InputStream somewhere in a BufferedInputStream.

src/main/java/minijava/MJError.java
```diff
@@ -0,0 +1,13 @@
+package minijava;
+
+/** Basic error class in this project. */
+public class MJError extends Error {
```
sherter: What's the rationale behind extending `Error` rather than `Exception` or `RuntimeException`?

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

src/main/java/minijava/lexer/BasicLexerInput.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public boolean hasNext() {
+    return currentChar != -1;
+  }
+
+  @Override
+  public Integer next() {
+    try {
+      currentChar = stream.read();
+      if (currentChar == '\n') {
```
sherter, line 34: This is probably simplified too much (what about, for example, `\r`?), but since error locations are only bonus I'm okay with it.

src/main/java/minijava/lexer/Token.java
```diff
@@ -0,0 +1,59 @@
...
+
+/** Lexer token */
+public class Token {
+
+  private final Terminal terminal;
+  private final Location location;
+  private final int content;
+  private final StringTable stringTable;
+
+  public Token(Terminal terminal, Location location, int content, StringTable stringTable) {
```
sherter: Generally, I like to keep things as hidden as possible (since only the Lexer creates Tokens, the constructor can be package-private, i.e remove the modifier). This applies to a few other places in this PR.

sherter (changes requested):
This is a very quick review. It contains some suggestions that came to my mind while quickly skimming through the code.
For the more important architectural aspects I need to take more time, but not today ;)

src/main/java/minijava/lexer/BasicLexerInput.java
parttimenerd: Good point.

src/main/java/minijava/MJError.java
parttimenerd: As far as I know, Exceptions have to be checked. That's a problem when using interfaces like `Iterator`.

src/main/java/minijava/lexer/Token.java
parttimenerd: Good point. I never really thought about this.

src/main/java/minijava/MJError.java
sgraf812: `RuntimeException`s are also unchecked.

parttimenerd: I didn't know this.

src/main/java/minijava/lexer/Token.java
sgraf812: Hmm... I don't think that's a good idea, the test suite will also have to create tokens.

src/main/java/minijava/MJError.java
sgraf812: No problem :)

src/main/java/minijava/lexer/BasicLexerInput.java
sgraf812: We can probably ignore `\r`: On windows it's immediately followed by `\n` and Mac has long ago switched from `\r` newlines to `\n`.

I like how this decomposes `Location`-generation into a separate `Iterator`, but not so much that it's stateful. I'll accept this as it is, but maybe at a later point we should separate `Location` generation from `current()/next(`)`stuff (e.g. input buffering), so that we can express`Location`generation as a fold and leave the complicated buffering details to a separate combinator. This way we can also get rid of the`lookAhead` function in the Lexer.

src/main/java/minijava/lexer/Lexer.java
```diff
@@ -0,0 +1,22 @@
...
+  /** Get the n.th next token. */
+  Token lookAhead(int lookAhead);
+
+  StringTable getStringTable();
+
+  default Stream<Token> stream() {
+    final Iterable<Token> iterable = () -> this;
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+}
```
sgraf812, line 22: (Dito about buffering and compositionality)

src/main/java/minijava/lexer/Token.java
```diff
@@ -0,0 +1,59 @@
+package minijava.lexer;
+
+/** Lexer token */
+public class Token {
+
+  private final Terminal terminal;
+  private final Location location;
+  private final int content;
+  private final StringTable stringTable;
```
sgraf812, line 9: We shouldn't carry around a `StringTable` reference in each `Token`, instead pass it to `getContentString()` as needed. It creates the false impression that we are actually modifying the `StringTable`, also memory pressure.

src/main/java/minijava/lexer/LexerInput.java
```diff
@@ -0,0 +1,23 @@
+package minijava.lexer;
+
+import java.util.Iterator;
+
+/** Input for a lexer (a stream of integers) it stops after the EOF character is read. */
+interface LexerInput extends Iterator<Integer> {
```
sgraf812: Is there a technical reason for `Integer` here? Why not `Byte`?

src/main/java/minijava/lexer/SimpleLexer.java
```diff
@@ -0,0 +1,410 @@
...
+    keywordTerms.put(StringTable.WHILE_KEYWORD_ID, Terminal.WHILE);
+    keywordTerms.put(StringTable.IF_KEYWORD_ID, Terminal.IF);
+    keywordTerms.put(StringTable.ELSE_KEYWORD_ID, Terminal.ELSE);
+    keywordTerms.put(StringTable.TRUE_KEYWORD_ID, Terminal.TRUE);
+    keywordTerms.put(StringTable.FALSE_KEYWORD_ID, Terminal.FALSE);
+    keywordTerms.put(StringTable.PUBLIC_KEYWORD_ID, Terminal.PUBLIC);
+    keywordTerms.put(StringTable.STATIC_KEYWORD_ID, Terminal.STATIC);
+    keywordTerms.put(StringTable.VOID_KEYWORD_ID, Terminal.VOID);
+    keywordTerms.put(StringTable.NULL_KEYWORD_ID, Terminal.NULL);
+    keywordTerms.put(StringTable.STRING_KEYWORD_ID, Terminal.STRING);
```
sgraf812, line 42: We _could_ move this work into a static initializer, however I don't think it's worth the trouble...

src/main/java/minijava/lexer/SimpleLexer.java
```diff
@@ -0,0 +1,410 @@
...
+    current = nextToken();
+    if (!lookAheadBuffer.isEmpty()) {
+      lookAheadBuffer.remove(0);
+    }
+    if (current.isEOF()) {
+      numberOfEOFs++;
+    }
+    return current;
+  }
+}
```
sgraf812, line 457: Some points:
1. As above, we should probably split off lookahead logic into a separate `Iterable`...
2. nextToken(), next(), lookAhead() and current() are really conflated and I'm not sure it's bug-free, but let's see it with tests
3. Also wasn't this supposed to be a Table-driven lexer?

Otherwise I'm fine with this.

parttimenerd: Some answers
1. i understand your point, but I think that it's to complicated for a simple Lexer.
2. Modifications are appreciated. But I think that the current implementation is atleast a working one.
3. Yes. But I got weird results with it. Writing a new lexer seemed to be much simpler to be than debugging the table based lexer.

src/main/java/minijava/lexer/LexerInput.java
parttimenerd: No (i just forgot to replace it with `Byte`).

src/main/java/minijava/lexer/Token.java
parttimenerd: The problem with your approach is that it will make debugging and parsing more complex. Token then can't implement a simple toString() method.

sgraf812: Ah, right.  Then I'm fine with this for the time being.

src/main/java/minijava/lexer/LexerInput.java
sgraf812: Maybe do this real quick and I'll accept.

parttimenerd: I'm working on it.

parttimenerd: I'm currently fuzzing the lexer. If the lexer doesn't yield to errors, then I commit it.

PR #18, opened 2016-10-26
Quickcheck the Lexer

sgraf812:
The `getContentString()` stream was always correct, but some `Terminal`s were wrong, probably due to copy and paste. This should work better now.


sherter:
I don't really understand all of the quickcheck specific things, but I enabled the debug output and as far as I can tell, it looks good! :+1:

sgraf812:
Yeah, correct generation of tokens was quite hairy...

PR #19, opened 2016-10-26
Test for lookAhead and bugfix

Hu3bl:
adds a unit test for the lookAhead method in SimpleLexer.
Additionally fixes a bug when lookAhead provides not the same tokes as next method.

### DIFF
--- a/src/main/java/minijava/lexer/SimpleLexer.java
+++ b/src/main/java/minijava/lexer/SimpleLexer.java
@@ -441,9 +441,6 @@ public class SimpleLexer implements Lexer {
   @Override
   public Token next() {
     current = nextToken();
-    if (!lookAheadBuffer.isEmpty()) {
-      lookAheadBuffer.remove(0);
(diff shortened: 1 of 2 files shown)


sherter:
@Hu3bl I was just about to submit my review, now it's to late :smile: 

I wanted to add that you should pass the Charset to `getBytes()` explicitly, in case someone uses a weird default system charset, say UTF-16 (is this even possible? I don't know, but make it explicit and we dont't have to worry about it :wink: )

parttimenerd:
That's only the test. The lexer throws a LexerError if a character is > 127.

sherter:
Then the test would be wrong... why wouldn't you want to fix bugs in tests?

parttimenerd:
Good point. But, as you already mentioned, it shouldn't be of concern.

Hu3bl:
I'm going to fix this @sherter. I will update it to bytes

PR #21, opened 2016-10-27
Implemented a LookAheadIterator, decoratin a regular Iterator

sgraf812:
Not ready to merge yet, because no tests. Also I plan to simplify stuff in `BasicLexerInput` and `SimpleLexer` with this. Also docs aren't really in sync any more.

### DIFF
--- a/src/main/java/minijava/lexer/SimpleLexer.java
+++ b/src/main/java/minijava/lexer/SimpleLexer.java
@@ -1,6 +1,5 @@
 package minijava.lexer;
 
-import static minijava.lexer.Terminal.INVERT;
 import static minijava.lexer.Terminal.RESERVED_OPERATORS;
 
 import java.io.ByteArrayInputStream;
--- a/src/main/java/minijava/lexer/SimpleLexer.java
+++ b/src/main/java/minijava/lexer/SimpleLexer.java
@@ -1,28 +1,30 @@
 package minijava.lexer;
 
+import static minijava.lexer.TaggedWithLocation.tagBytes;
 import static minijava.lexer.Terminal.RESERVED_OPERATORS;
 
 import java.io.ByteArrayInputStream;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import minijava.MJError;
+import minijava.util.InputStreamIterator;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /** SLL(1) parser style lexer implementation. */
 public class SimpleLexer implements Lexer {
 
-  private LexerInput input;
-  private StringTable stringTable = new StringTable();
-  private Token current = null;
-  private List<Character> characters = new ArrayList<>();
-  private Location location;
-  private List<Token> lookAheadBuffer = new ArrayList<>();
-  private Map<Integer, Terminal> keywordTerms = new HashMap<>();
-  private int numberOfEOFs = 0;
+  private static final List<Character> WHITESPACE = Arrays.asList(' ', '\n', '\r', '\t');
+  private final Iterator<TaggedWithLocation<Byte>> input;
+  private final StringTable stringTable = new StringTable();
+  private final Map<Integer, Terminal> keywordTerms = new HashMap<>();
+  /** Null if and only if either - We haven't yet called input.next() - input.hasNext() == false */
+  private TaggedWithLocation<Byte> current;
+  /** Null if and only if we haven't lexed the next token yet. See hasNext. */
+  private Token leftOver;
 
-  public SimpleLexer(LexerInput input) {
-    this.input = input;
+  public SimpleLexer(Iterator<Byte> input) {
+    this.input = tagBytes(input);
     keywordTerms.put(StringTable.BOOLEAN_KEYWORD_ID, Terminal.BOOLEAN);
     keywordTerms.put(StringTable.INT_KEYWORD_ID, Terminal.INT);
     keywordTerms.put(StringTable.CLASS_KEYWORD_ID, Terminal.CLASS);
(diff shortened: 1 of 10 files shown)


src/main/java/minijava/util/LookAheadIterator.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Queue;
+
+/**
+ * An @Iterator@ decorator which enables look ahead into iterated elements.
+ */
+public class LookAheadIterator<T> implements Iterator<T> {
+
+    private final Iterator<T> it;
+    private final ArrayDeque<T> nextElements = new ArrayDeque<>(16);
+
```
parttimenerd, line 13: Why do you limit the possible number of lookahead tokens?

src/main/java/minijava/util/LookAheadIterator.java
```diff
@@ -0,0 +1,81 @@
...
+        if (nextElements.isEmpty() & n == 0) {
+            // This can only happen if next() hasn't been called yet.
+            // Otherwise index 0 in nextElements would be filled with the current element.
+            // In this case the current element (index 0) doesn't exist!
+            return Optional.empty();
+        }
+
+        while (nextElements.size() <= n) {
+            if (!it.hasNext()) {
+                return Optional.empty();
```
parttimenerd: It would be nicer to return some sort of end element (i.e. the last token of the stream oder the `eof` character.

parttimenerd (commented):
Nice abtraction that is more elegant than the current solution. But I don't think that we need for the LexerInput.

src/main/java/minijava/util/LookAheadIterator.java
sgraf812: That's just the initial capacity, no?

src/main/java/minijava/util/LookAheadIterator.java
sgraf812: Don't think so. You could just use [`Optional.orElse`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElse-T-) for that.

src/main/java/minijava/util/LookAheadIterator.java
parttimenerd: You're correct. I confused with a buffer class.

src/main/java/minijava/util/InputStreamIterator.java
```diff
@@ -0,0 +1,33 @@
...
+      return input.available() > 0;
+    } catch (IOException e) {
+      return false;
+    }
+  }
+
+  @Override
+  public Byte next() {
+    try {
+      return (byte) input.read();
```
parttimenerd, line 28: Please add a check for values greater than 127 here (as they are not supported).

src/main/java/minijava/lexer/TaggedWithLocation.java
```diff
@@ -0,0 +1,42 @@
+package minijava.lexer;
+
+import java.util.Iterator;
+
+public class TaggedWithLocation<T> {
+
```
parttimenerd, line 6: Should this replace the location field in the Token class?

src/main/java/minijava/lexer/TaggedWithLocation.java
```diff
@@ -0,0 +1,42 @@
...
+        byte b = bytes.next();
+
+        if (b == '\n') {
+          currentLine++;
+          currentColumn = 0;
+        } else {
+          currentColumn++;
+        }
+
+        return new TaggedWithLocation<>(l, b);
```
parttimenerd, line 38: You assign a location to every byte?

parttimenerd (commented):
I get that this my be, from a functional programming standpoint, cleaner and easier to understand.
But I don't think that it's better or more comprehensible than the oid code.

src/main/java/minijava/util/InputStreamIterator.java
sgraf812: That's done in the Lexer, included in the check for `<= 0` (`byte` is signed).

src/main/java/minijava/lexer/TaggedWithLocation.java
sgraf812: Yep... Just throwing stuff at the wall and see what sticks... No need to review at this point.

src/main/java/minijava/lexer/SimpleLexer.java
```diff
@@ -41,336 +43,476 @@ public SimpleLexer(LexerInput input) {
...
+
+    while (WHITESPACE.contains((char) (byte) current.tagged)) {
+      if (!tryAdvance()) return null;
+    }
+
+    assert current != null;
+
+    byte cur = current.tagged;
+    Location location = current.loc;
+
```
parttimenerd, line 106: Is there a reason why you create a location object for each byte?

sgraf812:
So all in all, things got a little slower while the code is a little clearer I think (although clarity is certainly in the eye of the beholder ;)).

Pros of this approach:
- Separate concerns/compositionality, which enables easier testing (e.g. splitting off `LookAheadIterator`, `InputStreamIterator`, `TaggedWithLocation`). Although we are already pretty good with this point (well, apart from the fact that `lookAhead` on master is unused until now)
- IMHO clearer code/invariants in `SimpleLexer` compared to `master` as a result. I couldn't keep in mind all the different assumptions/calling relations of `next()`, `hasNext()`, `lookAhead`, `lookAhead()`, `nextToken`, `current`, etc.

Cons:
- Quite involved refactoring, seemingly unnecessary at this point (but who knows what kind of bugs might hide in some corner case)
- Different semantics: `BasicLexer` doesn't emit `EOF` anymore. Although that can be regained by concatenating with an infinite `Iterator` always emitting `EOF`, or using simply using `LookAheadIterator` with `input.lookAhead(n).orElse(Terminal.EOF)` in Parser code. IMO `EOF` is in-band signaling, overlapping with `hasNext()`, which doesn't really fit the Iterator pattern...
- A little slower (done no benchmarks), but acceptably fast still (The quickcheck tests went up from 1.5 to 1.7 s for me).

src/main/java/minijava/lexer/SimpleLexer.java
sgraf812: Nope, but doesn't hurt AFAICT.

sherter:
First of all, thanks for your effort!
To be honest, I find both the existing and your approach rather complex. Don't get me wrong, usually I love good abstractions and composability and stuff, but I don't think we gain very much from it at this point.
If I have to look at this code again in 4 weeks to fix some nasty bug, I would rather look at a single class (for example, like in the [go lexer](https://github.com/golang/go/blob/master/src/go/scanner/scanner.go)) than at our code.

Maybe we should focus more on the high-level interface in the future and agree on that, before someone presents an implementation, at which point we're pretty much stuck with it. I would actually prefer to go with APIs similar to what they presented us in the lecture. It might not be the "best" (whatever that means) approach, but at least everybody understands it (hopefully), and that is what "clarity" is, isn't it?

sgraf812:
Agreed. I'll freeze this.

PR #23, opened 2016-10-29
Refactor CLI and corresponding tests to be more straight forward

sherter:
I tried to clarify and simplify things. The ugly parts (that deal with
the JCommander library) were moved into a separate class.

Blackbox tests that use the ./run script were removed. We have the
python based test suites for that. Some other tests for command line
options validation were added.

I'm not sure why the outputLexerTokens method was so complicated before,
I hope I didn't miss something?!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,8 @@ dependencies {
     testCompile 'com.pholser:junit-quickcheck-core:0.6.1'
     testCompile 'com.pholser:junit-quickcheck-generators:0.6.1'
     testCompile 'junit:junit:4.12'
+    testCompile 'org.hamcrest:hamcrest-library:1.3'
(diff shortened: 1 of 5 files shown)


sgraf812:
Good stuff! Extracting tests going through the run script means that tests will go through on Windows again (totally living in the moment :pray:).

Although I don't get the `FileSystem` stuff. Is it some kind of virtualization for easier testing? Or just a more convenient abstraction than the default Java one?

sherter:
Right, injecting the FileSystem makes testing easier. `Paths.get()` really is `FileSystems.getDefault().getPath()` and by injecting the FileSystem, I can use an in memory FileSystem in the tests (https://github.com/google/jimfs). It is a Java 7 abstraction...

PR #25, opened 2016-11-01
recursive descent parser draft

Hu3bl:
Draft of the parser. it is not tested yet.

### DIFF
--- a/src/main/java/minijava/parser/Parser.java
+++ b/src/main/java/minijava/parser/Parser.java
@@ -1,8 +1,5 @@
 package minijava.parser;
 
-import static minijava.token.Terminal.INT;
-import static minijava.token.Terminal.SEMICOLON;
-
 import minijava.lexer.Lexer;
 import minijava.token.Terminal;
 import minijava.token.Token;
@@ -36,6 +33,13 @@ public class Parser {
   private boolean isCurrentTokenNotTypeOf(Terminal terminal) {
     return !isCurrentTokenTypeOf(terminal);
   }
+
+    private boolean isCurrentTokenBinaryOperator() {
+        if(currentToken.isType(Terminal.TerminalType.OPERATOR)) {
+            return true;
+        }
+        return false;
+    }
 
   public void parse() {
     consumeToken();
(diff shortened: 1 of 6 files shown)


src/main/java/minijava/parser/Parser.java
```diff
@@ -0,0 +1,450 @@
...
+      parseExpressionWithPrecedenceClimbing(precedence);
+    }
+  }
+
+  /** UnaryExpression -> PostfixExpression | (! | -) UnaryExpression */
+  private void parseUnaryExpression() {
+    switch (currentToken.terminal) {
+      case INVERT:
+      case MINUS:
+        parseUnaryExpression();
```
parttimenerd, line 309: Here's probably a consume action missing.

src/main/java/minijava/parser/Parser.java
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  /** BlockStatement -> Statement | LocalVariableDeclarationStatement */
+  private void parseBlockStatement() {
+    switch (currentToken.terminal) {
+      case INT:
+      case BOOLEAN:
+      case VOID:
+      case IDENT:
+        parseLocalVariableDeclarationStatement();
```
parttimenerd: Consider the statement "a = 3;" or "a[4] = 5;" both wont match this rule.

src/main/java/minijava/parser/Parser.java
Hu3bl: You're right, expectAndConsume is missing

parttimenerd: I'm currently fixing it

src/main/java/minijava/parser/Parser.java
Hu3bl: The problem is IDENT, isn't it? Your statements need to be parsed as Statement and not as LocalVariableDeclarationStatement, because it is not a declaration.

parttimenerd: Yes, but I fixed the problem (commits are coming soon).

PR #27, opened 2016-11-02
Fuzzy testing the Parser

sgraf812:
I extracted generation of the token stream into a separate class, so we can maybe also use it in `mjtest`. 

Right now the tests fail, but I haven't figured out yet if it's the tests or the parser.

### DIFF
--- a/src/main/java/minijava/parser/Parser.java
+++ b/src/main/java/minijava/parser/Parser.java
@@ -18,7 +18,11 @@ public class Parser {
   }
 
   private void consumeToken() {
-    this.currentToken = tokens.next();
+    if (tokens.hasNext()) {
+      this.currentToken = tokens.next();
+    } else {
+      this.currentToken = EOF_TOKEN;
+    }
   }
 
   private void expectAndConsume(Terminal terminal) {
@@ -156,8 +160,9 @@ public class Parser {
   /** Parameters -> Parameter | Parameter , Parameters */
   private void parseParameters() {
     parseParameter();
-    if (isCurrentTokenTypeOf(COMMA)) {
-      parseParameters();
+    while (isCurrentTokenTypeOf(COMMA)) {
+      expectAndConsume(COMMA);
+      parseParameter();
     }
   }
 
@@ -233,7 +238,8 @@ public class Parser {
   /** BlockStatement -> Statement | LocalVariableDeclarationStatement */
   private void parseBlockStatement() {
     if (currentToken.isOneOf(INT, BOOLEAN, VOID)
-        || matchCurrentAndLookAhead(IDENT, LBRACKET, RBRACKET)) {
+        || matchCurrentAndLookAhead(IDENT, LBRACKET, RBRACKET)
+        || matchCurrentAndLookAhead(IDENT, IDENT)) {
       parseLocalVariableDeclarationStatement();
     } else {
       parseStatement();
@@ -375,6 +381,7 @@ public class Parser {
     if (isCurrentTokenNotTypeOf(RBRACKET)) {
       parseExpression();
       while (isCurrentTokenTypeOf(COMMA) && isCurrentTokenNotTypeOf(EOF)) {
+        expectAndConsume(COMMA);
         parseExpression();
       }
     }
@@ -461,7 +468,7 @@ public class Parser {
     expectAndConsume(LBRACKET);
     parseExpression();
     expectAndConsume(RBRACKET);
-    while (isCurrentTokenTypeOf(LBRACKET) && isCurrentTokenNotTypeOf(EOF)) {
+    while (matchCurrentAndLookAhead(LBRACKET, RBRACKET)) {
       expectAndConsume(LBRACKET);
       expectAndConsume(RBRACKET);
     }
--- a/src/main/java/minijava/parser/Parser.java
+++ b/src/main/java/minijava/parser/Parser.java
@@ -174,6 +174,8 @@ public class Parser {
 
   /** Type -> BasicType ([])* */
   private void parseType() {
+    // Only later call is in parseLocalVariableDeclarationStatement()
+    // parseType() does not recurse however, so we are safe.
     parseBasicType();
     while (isCurrentTokenTypeOf(LBRACKET) && isCurrentTokenNotTypeOf(EOF)) {
       expectAndConsume(LBRACKET);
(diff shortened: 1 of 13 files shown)


sgraf812:
I made a small refactoring in the `Parser`, substituting `Lexer` for `Iterable<Token>`, which I'm unable to merge now, because of `lookAhead` in `matchCurrentAndLookAhead`.

Is there a way to get rid of that? Otherwise, I can offer to include `LookAheadIterator` once again, because I can't really test through the `Lexer`, because I specify the token stream directly.

parttimenerd:
Why don't you just create a sub class of Lexer?

sgraf812:
That way I have to duplicate the `lookAhead` logic.

sherter:
If we _do_ use the `Iterator<Token>` abstraction for our Lexer in the Parser, I think we should be consequent and remove the `EOF` token. I think somebody mentioned that before. Iterator and EOF doesn't really work together.

sgraf812:
Yeah, I plan to do these kinds of refactorings in a follow up PR.

Although getting rid of `EOF` also means that `hasNext()` might have to do actual lexing, since we can't just return `EOF` when there are only whitespache/comments left.

That was where most of the complexity came from in the ambitious refactoring I did a week ago.

src/main/java/minijava/parser/Parser.java
```diff
@@ -153,8 +160,9 @@ private void parseMethod() {
   /** Parameters -> Parameter | Parameter , Parameters */
   private void parseParameters() {
     parseParameter();
-    if (isCurrentTokenTypeOf(COMMA)) {
-      parseParameters();
+    while (isCurrentTokenTypeOf(COMMA)) {
+      expectAndConsume(COMMA);
+      parseParameter();
```
sgraf812, line 165: I figured that the bugs I found and fixed demand some explanation.

This one is the one from the other PR: Although Parameterlists of length > 1 in method prototypes are checked for with `COMMA`, the actual `COMMA` token isn't consumed, so  `IDENT COMMA IDENT` would complain that `COMMA` isn't an `IDENT`.

src/main/java/minijava/parser/Parser.java
```diff
@@ -230,7 +238,8 @@ private void parseBlock() {
   /** BlockStatement -> Statement | LocalVariableDeclarationStatement */
   private void parseBlockStatement() {
     if (currentToken.isOneOf(INT, BOOLEAN, VOID)
-        || matchCurrentAndLookAhead(IDENT, LBRACKET, RBRACKET)) {
+        || matchCurrentAndLookAhead(IDENT, LBRACKET, RBRACKET)
+        || matchCurrentAndLookAhead(IDENT, IDENT)) {
```
sgraf812, line 251: Obviously this wouldn't have matched stuff like `String s = "";`, because `String` is an `IDENT`, but `s` is also an `IDENT` and not an `LBRACKET`.

src/main/java/minijava/parser/Parser.java
```diff
@@ -372,6 +381,7 @@ private void parseArguments() {
     if (isCurrentTokenNotTypeOf(RBRACKET)) {
       parseExpression();
       while (isCurrentTokenTypeOf(COMMA) && isCurrentTokenNotTypeOf(EOF)) {
+        expectAndConsume(COMMA);
         parseExpression();
```
sgraf812, line 390: Same situation as with parameter lists, but here with argument lists.

src/main/java/minijava/parser/Parser.java
sgraf812: Also I got rid of the tail call, which, as we experienced, the JVM handles quite poorly...

src/main/java/minijava/parser/Parser.java
```diff
@@ -458,7 +468,7 @@ private void parseNewArrayExpression() {
     expectAndConsume(LBRACKET);
     parseExpression();
     expectAndConsume(RBRACKET);
-    while (isCurrentTokenTypeOf(LBRACKET) && isCurrentTokenNotTypeOf(EOF)) {
+    while (matchCurrentAndLookAhead(LBRACKET, RBRACKET)) {
```
sgraf812, line 476: In `new int[5][3]` we got a NewArrayExpression `new int[5]` followed by an IndexExpression `[3]` (or whatever it's called), which previously the code would try to interpret as a NewArrayExpression `new int[5][]` only to complain at the last `RBRACKET`.

src/main/java/minijava/parser/Parser.java
```diff
@@ -319,16 +330,10 @@ private void parseExpressionWithPrecedenceClimbing(int minPrecedence) {
...
-        consumeToken();
-        parseUnaryExpression();
-        break;
-      default:
-        parsePostfixExpression();
-        break;
+    while (currentToken.isOneOf(INVERT, MINUS)) {
+      consumeToken();
     }
+    parsePostfixExpression();
```
sgraf812, line 336: Also eliminated this tail call, inspired by PR #28. Also made sure, that this was the last recursion we could eliminate, all other recursion doesn't happen in tail position.

src/main/java/minijava/util/LookAheadIterator.java
```diff
@@ -0,0 +1,94 @@
...
+   *     element in the stream, respecting @Iterable@ semantics.
+   * @lookAhead(1)@ will point to the first element in that case.
+   *     <p>To fullfill calls to @lookAhead(n)@ with n greater than 1, it will call @next@ on the
+   *     underlying iterable as needed. A call to @lookAhead(0)@ is guaranteed not to call @next@.
+   * @param n non-negative integer, denoting the number of elements to look ahead.
+   * @return The element which will be the return value after @n-1@ calls to @next@.
+   */
+  public Optional<T> lookAhead(int n) {
+    if (n < 0) {
+      throw new IllegalArgumentException("n was negative");
```
sherter, line 37: Since we already have Guava in our project, you could use [checkArgument(n < 0, "n was negative")](https://github.com/google/guava/wiki/PreconditionsExplained) here, which would be slightly more elegant (only a suggestion for the future :wink: )

sherter:
I didn't review the code in `src/test` very thoroughly, but the changes in the production code look good!

src/main/java/minijava/util/LookAheadIterator.java
sgraf812: Thanks for the pointer!

PR #28, opened 2016-11-02
Fix "more than one parameter" parser bug

parttimenerd:


src/main/java/minijava/parser/Parser.java
```diff
@@ -310,16 +311,10 @@ private void parseExpressionWithPrecedenceClimbing(int minPrecedence) {
 
   /** UnaryExpression -> PostfixExpression | (! | -) UnaryExpression */
```
sherter, line 312: Could you add a test case that demonstrates the problem? Looking at the production, the existing code looks better than the new code ;) (Maybe the comment needs to change, too?)

parttimenerd: You're correct. I thought that the (unecessary) recursion could become a problem in tests with lots of `!` in a row. But this isn't actually the case (newer test case shows this).

sgraf812:
I already fixed this in the `parser-qcheck` branch. So, if we can discuss whether to merge that first, there will be less bugs we fix independently in the long rung.

parttimenerd:
We can just use you're bug fix in the `parse-qcheck` branch and delete this PR.

PR #29, opened 2016-11-03
Added mjtest as a submodule tracking master

sgraf812:
Also modified build.gradle so that it will invoke mjt.py ast and syntax tests as part of `gradle check`.

The tests currently don't exit successfully, so we should probably wait until the fixes in the parser land.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "mjtest"]
+	path = mjtest
+	url = https://github.com/mj3-16/mjtest
+	branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
 language: java
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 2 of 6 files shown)


sgraf812:
OK, after some fiddling and `git push -f` (sorry if I messed up someones local tracking branch), there are 6 currently failing tests left:
1. `if () { }` invalid
2. `while () { }` invalid
3. `int x = [];` invalid
4. `SomeValue[] val = new SomeValue[];` invalid
5. `/* some special chars for you:   */` lexer error because of `\0`, which should be valid in a comment according to spec (🙄)
6. `return a((a((a((a((a((a((a((a((a((a((a((a((a(())))))))))))))))))))))))));` invalid (inner () in method call, related to 1. and 2.: empty expressions)

These are mostly testing invalid files, which wasn't possible with the fuzzy tester, so this is good!

I'd say, fix these things on another branch and open PRs against `mjtest-submodule`. Then we can quickly see if a certain fix is actually a fix.

PR #30, opened 2016-11-03
Fix stackoverflow in Lexer

sherter:


sgraf812:
Something I realized just yesterday evening: The simplest fix would be to wrap everything in `parseNextToken()` in a `while (true) { .. }` and in the `/` case instead of recursing do `continue;` or just fall through. This would work since every other case just `return`s immediately.

sherter:
Wrapping everything in a while loop is basically what I did in [L418](https://github.com/mj3-16/minijavac/pull/30/files#diff-8ce91d7f20d797a74e971a71f1230cb4R418) (except for making it explicit that we skip `COMMENT`s).
I could agree to blowing up the `parseNextToken` method itself only if we also delete the `COMMENT` token, because then it would really be obsolete...

sgraf812:
You're right. Let's discuss this on slack: I don't think we need a `COMMENT` terminal anymore.

sgraf812:
Although we should get rid of `WS` in the same step then. I'd say, we can merge this for now and leave that open for a separate PR: `COMMENT`, `WS` and `EOF` should all be removed IMO.

PR #32, opened 2016-11-04
Bumped mjtest

sgraf812:
Some test failures:

1. decision_between_localVarDecl_and_statement.mj: Fails, but rightly so: `public int b = 0;` is invalid
2. empty_statements.mj: `int field1 = 0;` also invalid
3. new_array_expression.mj: Same
4. parsing_recursion.mj: Long sequence of `true||(true||...`. Have to investigate if this really should not overflow. That might not even be the issue here: It's just wrong syntax again; `while` is not allowed in the body of a class.
5. simple_class.mj: Invalid syntax: `while` in class body

So all of them are actually errors in the test suite.


sgraf812:
After another bump (`--amend`ed again), this looks fine.

PR #34, opened 2016-11-04
Fix "'new Object;' should be invalid" bug

parttimenerd:


sgraf812:
Could you also bump the submodule? Otherwise the new tests aren't included.

src/main/java/minijava/parser/Parser.java
```diff
@@ -467,6 +468,8 @@ private void parseNewObjectArrayExpression() {
           case LBRACKET:
             parseNewArrayExpression();
             break;
+          default:
+            unexpectCurrentToken();
```
sgraf812: Maybe change `expectAndConsume` to take var args, so that error messages can be better? e.g. `expectAndConsume(LBRACKET, LPAREN)`

sgraf812: Ah, we're using this quite often. Maybe we should address this separately, also quality of error messages isn't paramount.

sgraf812 (approved):

src/main/java/minijava/parser/Parser.java
parttimenerd: I*ve already changed it.

sgraf812:
Also, you will need to reformat your code. I have the feeling that the git hook isn't working for you in the same way it isn't for me: it will format the code, but won't `git add` the formatted code afterwards.

sgraf812 (approved):
<3

PR #38, opened 2016-11-04
Simplified the Lexer

sgraf812:
Got rid of some redundant state. Also fixed the git hook to cope with renames.

### DIFF
--- a/src/main/java/minijava/lexer/Lexer.java
+++ b/src/main/java/minijava/lexer/Lexer.java
@@ -408,7 +408,7 @@ public class Lexer implements Iterator<Token> {
 
   @Override
   public boolean hasNext() {
-    return token != null && token.isEOF();
+    return token == null || !token.isEOF();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


sgraf812:
Uhmm.. that probably was one removed redundancy too much...

src/main/java/minijava/lexer/Lexer.java
```diff
@@ -409,34 +406,24 @@ private Token createToken(Terminal terminal, String content) {
...
-    } else {
-      current = parseNextToken();
-    }
-    return current;
-  }
-
   @Override
   public boolean hasNext() {
-    return numberOfEOFs < 1;
+    return token == null || !token.isEOF();
```
sherter, line 411: `token == null` is only true once (in the beginning). _**If**_ we were optimizing for performance, the test for `!token.isEOF()` should probably be the first one :stuck_out_tongue:

sgraf812: Even if, I'm inclined to say this won't matter.

sherter: agreed :wink:

PR #39, opened 2016-11-05
Refactor terminal - a first attempt

sherter:
Please focus on the differences in the `Terminal` and `Token` class and tell me what you think. I didn't fix all the tests just yet, in case you don't like the proposal.

### DIFF
--- a/src/main/java/minijava/token/Terminal.java
+++ b/src/main/java/minijava/token/Terminal.java
@@ -1,114 +1,92 @@
 package minijava.token;
 
-import static minijava.token.Terminal.TerminalType.CONTROL_FLOW;
-import static minijava.token.Terminal.TerminalType.HIDDEN;
-import static minijava.token.Terminal.TerminalType.LITERAL;
-import static minijava.token.Terminal.TerminalType.MISC;
-import static minijava.token.Terminal.TerminalType.OPERATOR;
-import static minijava.token.Terminal.TerminalType.SYNTAX_ELEMENT;
-import static minijava.token.Terminal.TerminalType.TYPE;
+import static minijava.token.Terminal.Associativity.LEFT;
+import static minijava.token.Terminal.Associativity.RIGHT;
+
+import java.util.Optional;
 
 /** Enum of terminals used by the Lexer. */
 public enum Terminal {
-  EOF("eof", MISC),
-  COMMENT("comment", HIDDEN),
-  WS("white space", HIDDEN),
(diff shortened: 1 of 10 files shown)


src/main/java/minijava/token/Terminal.java
```diff
@@ -1,114 +1,92 @@
...
+    LEFT,
+    RIGHT
   }
 
-  public TerminalType getType() {
-    return terminalType;
-  }
+  public final Optional<String> string;
+  final Optional<Associativity> associativity;
+  final Optional<Integer> precedence;
```
sgraf812: Well, the public field is OK I guess, but I found this rather interesting: http://blog.joda.org/2015/08/java-se-8-optional-pragmatic-approach.html

TLDR; Use `Optional` only as return value from public APIs, otherwise use `null` (and document so).

sgraf812 (approved):
I like where this is going

src/main/java/minijava/token/Terminal.java
sherter: Looks much better now :+1:

src/main/java/minijava/token/Token.java
```diff
@@ -1,50 +1,59 @@
...
+  public boolean isOperator() {
+    return terminal.associativity != null;
   }
 
-  public boolean isType(Terminal.TerminalType type) {
-    return terminal.isType(type);
-  }
-
-  @Override
-  public String toString() {
```
parttimenerd, line 29: It might be useful to have a function in the refactored Token class that outputs the same information for error messages.

parttimenerd:
I like the current refactorings too.

sgraf812:
I could help with updating `SimpleLexerProperties` (missed to rename it to `LexerProperties`), but I haven't really time for the rest of the weekend.

sherter:
I'm working on it (it's a little tricky though). Maybe I'll remind you of that offer, if I can't figure it out :wink:

sherter:
Whitespace and comments are currently not inserted randomly, because the corresponding tokens were removed. Otherwise I think this is ready for merging.

src/main/java/minijava/token/Token.java
sherter: I'm not sure what you mean by that. Do you need a method that outputs something different from the `toString` method?

parttimenerd: Yes, to simplify the creation of error messages.
But it doesn't matter in this context (we can just change der ParserError class).

PR #41, opened 2016-11-05
Get rid of LexerInput

sherter:
Also rename methods from `parseX` to `scanX` in the lexer (to avoid confusion with the `parseX` methods in the parser).


sgraf812:
Hmm... I think this is better overall, but I don't like two things:
1. Doing location calculation, low-level `InputStream.read()` and lexing in the same class seems bad from the perspective of separation of concerns
2. ~~Taking an `InputStream` as constructor parameter feels meh, much rather like `Iterator<Byte>` (although that's a concern also with the old approach).~~ Just recognized the constructor overloads, I'm OK with that

Edit: Whatever, vote for merge

PR #44, opened 2016-11-10
Parser with AST construction

Hu3bl:
Draft of parser, which constructs the abstract syntax tree.
Parser is tested, constructed AST is not. How do we do this? With the pretty printer?


sgraf812:
Yep, I'd say so. I'll hack some property-based tests together tomorrow.

PR #50, opened 2016-11-13
Bumped mjtest, which is currently failing

sgraf812:
`mjtest` currently passes the wrong flag to `run`. It should be `--print-ast` instead of `--pretty-print`.


parttimenerd:
We should disable the format check for mjtest test cases…

PR #55, opened 2016-11-15
Located ast

sgraf812:
Again, mostly boiler plate. Java is just rediculous in this regard.

I haven't yet integrated proper squiggly underlines like [here](http://elm-lang.org/blog/compilers-as-assistants), but that will happen some time soon and is completely independent of this week's tasks.

We should probably catch all compilation-related exception at some level where we have access to the input string, otherwise underlining is not possible.

Also note that `SourceRange` has inclusive `begin` and exclusive `end`, as it was the sanest to use. Underlining would of course respect that.


src/main/java/minijava/ast/Statement.java
```diff
@@ -84,7 +107,7 @@ public ExpressionStatement(Expression<TRef> expression) {
 
     TRet visitBlock(Block<? extends TRef> that);
 
-    TRet visitEmptyStatement(EmptyStatement<? extends TRef> that);
+    TRet visitEmptyStatement(Empty<? extends TRef> that);
```
sherter: `visitEmpty(...)`

src/main/java/minijava/ast/Expression.java
```diff
@@ -198,24 +225,24 @@ public VariableExpression(TRef var) {
...
+    TReturn visitFieldAccess(FieldAccess<? extends TRef> that);
 
-    TReturn visitArrayAccess(ArrayAccessExpression<? extends TRef> that);
+    TReturn visitArrayAccess(ArrayAccess<? extends TRef> that);
 
-    TReturn visitNewObjectExpr(NewObjectExpression<? extends TRef> that);
+    TReturn visitNewObjectExpr(NewObject<? extends TRef> that);
 
-    TReturn visitNewArrayExpr(NewArrayExpression<? extends TRef> size);
+    TReturn visitNewArrayExpr(NewArray<? extends TRef> size);
```
sherter: `visitNewArray(...)`

src/main/java/minijava/ast/Expression.java
```diff
@@ -198,24 +225,24 @@ public VariableExpression(TRef var) {
...
+    TReturn visitMethodCall(MethodCall<? extends TRef> that);
 
-    TReturn visitFieldAccess(FieldAccessExpression<? extends TRef> that);
+    TReturn visitFieldAccess(FieldAccess<? extends TRef> that);
 
-    TReturn visitArrayAccess(ArrayAccessExpression<? extends TRef> that);
+    TReturn visitArrayAccess(ArrayAccess<? extends TRef> that);
 
-    TReturn visitNewObjectExpr(NewObjectExpression<? extends TRef> that);
+    TReturn visitNewObjectExpr(NewObject<? extends TRef> that);
```
sherter: `visitNewObject(...)`

sherter (changes requested):
Some ideas for small adjustments for more consistency...

sherter (approved):

PR #56, opened 2016-11-15
Add Ref for name analysis and add Nameable interface

sherter:
Add Nameable interface, so we don't have to rely on Object's toString() method. Also add `Definition` and `Ref` classes for name analysis.


sherter:
Superseded by #59 .

PR #57, opened 2016-11-15
Format fix

sherter:
Verification for proper formating fails, because we use `SNAPSHOT` versions of google-java-format. Out local version and the CI version probably became out of sync. We were using the 1.2-SNAPSHOT version because of better windows support, as I remember. Maybe our only windows user (I won't mention your name :wink: ) can manage to work with `\n` instead of `\r\n`?


sgraf812:
That's racist :sob: But I can probably configure IntelliJ so that it won't mess up line separators.

PR #58, opened 2016-11-15
Add a .gitattributes file stating that every text file is included with LF as line separator

sgraf812:


sherter:
I don't really know how gitattributes work, but maybe `*` is a little dangerous because of our test files? We don't want git to mess around with them...

sgraf812:
I don't think git will touch anything in a submodule.

PR #76, opened 2016-11-27
Graph building visitor

sgraf812:
This weeks efforts


src/main/java/minijava/Cli.java
```diff
@@ -29,14 +30,16 @@
...
+                "  --echo          write file's content to stdout",
+                "  --lextest       run lexical analysis on file's content and print tokens to stdout",
+                "  --parsetest     run syntactical analysis on file's content",
+                "  --print-ast     pretty-print abstract syntax tree to stdout",
+                "  --check         parse the given file and perform semantic checks",
+                "  --compile-firm  compile the given file with the libfirm amd64 backend",
+                "  --run-firm      compile and run the given file with libfirm amd64 backend",
+                "  --help          display this help and exit",
                 "",
                 "  One (and only one) of --echo, --lextest, --parsetest or --print-ast is required."
```
sherter, line 44: New options are missing. Probably better to stop listing them all here...

src/main/java/minijava/ir/IREmitter.java
```diff
@@ -0,0 +1,769 @@
...
+   */
+  @Nullable private Function<Node, Node> storeInCurrentLval;
+
+  public static void main(String[] main_args) {}
+
+  @Override
+  public Void visitProgram(minijava.ast.Program that) {
+    classTypes.clear();
+    methods.clear();
+    fields.clear();
```
sherter, line 98: If you intend to make this Visitor reusable, you probably need to clear some more state. These things are easily forgotten, maybe it's better to just document that this visitor can only be used once?

src/main/java/minijava/ir/IREmitter.java
```diff
@@ -0,0 +1,769 @@
...
+   * function for getting the address of a local variable as a Node. So in order to not duplicate a
+   * lot of work (e.g. computing array offsets, etc.) in a mechanism without this variable, we
+   * abstract actual assignment out into a function.
+   */
+  @Nullable private Function<Node, Node> storeInCurrentLval;
+
+  public static void main(String[] main_args) {}
+
+  @Override
+  public Void visitProgram(minijava.ast.Program that) {
```
sherter, line 95: In general: I would prefer to use simple names for our own classes and fully qualified ones for the firm classes (they are also shorter since everything is in the `firm` package)

src/main/java/minijava/ir/IREmitter.java
```diff
@@ -0,0 +1,769 @@
...
+  private Entity addFieldDecl(Field f) {
+    Type type = getStorageType(f.type);
+    ClassType definingClass = classTypes.get(f.definingClass.def);
+    Entity fieldEnt = new Entity(definingClass, f.name(), type);
+    fieldEnt.setLdIdent(NameMangler.mangleInstanceFieldName(definingClass.getName(), f.name()));
+    return fieldEnt;
+  }
+
+  /**
+   * Returns the primitive type in which to store type @type@. Consider a @boolean b;@ declaration,
```
sherter, line 135: I have seen the `@ @` markup a few times in our code now. But that isn't valid javadoc markup, is it? Maybe use `<em></em>` or `{@code }` instead?

sherter (commented):
Just a few general notes. For more in depth analysis I need to get my hands dirty - which it is too late for this night ;)

src/main/java/minijava/ir/IREmitter.java
sgraf812: I would be fine with that.

src/main/java/minijava/ir/IREmitter.java
sgraf812: This isn't documented, but those three maps are the 'per program' state, while the other state is reset per method declaration (`construction`, `graph`, `localVarIndexes`) and per expression node (`storeInCurrentLval`).

Anyway, I'd say that you shouldn't use this visitor more than once, or at least not in parallel.

That's why I tend to interpret `Visitor`s not as an external interface, but rather as a convenience for implementing tree traversal. So I would expose a static method like `compile` which would create a new instance of the `Visitor` any time it is called.

IMO `ast.acceptVisitor(new SemanticAnalyzer())` is also less expressive than `Semantics.analyze(ast)`, but YMMV.

So, I don't think we can get rid of the state so easily. I mean yeah, we could abstract it all into a private `VisitorState` class and create a fresh instance on every new `visitProgram`, but that's rather the same thing we're doing here, plus we'd still have to reset inidividual state in `emitBody` and expression visitors.

src/main/java/minijava/ir/IREmitter.java
sgraf812: Guilty, I thought Javadoc was less clumsy. Should read up on that, but I think I just use backticks rather than worsening my RSI by typing that kind of abomination.

sgraf812:
Well, with the proper inclusion of the `libfirm` and `jFirm` versions we should use, the diff has finally gone unbearable. For anyone who wants to see a more usable diff, use the commit range https://github.com/mj3-16/minijavac/compare/4ea86760aafecaaa0695521973c5dc82ca9b9622...d9583c0b63a2174ea169d16b8d8aab0d6b421a6e

src/main/java/minijava/ir/IREmitter.java
parttimenerd: There are already such static methods. We should probably make the constructor private.

parttimenerd:
Why can't we just keep the submodules and checkout the correct versions?

sgraf812:
Find me the correct SHAs and I'll do it. But e.g. `libfirm-1.22.0` is from 2012, whereas the source version we were given was `libfirm-1.22.1` and much more recent. Also, `jFirm`s last release was compatible with 1.21.0, so go figure.

And staying on `master` on both libraries is a bad idea. I don't want to have to switch to tarball versions every time I encounter a bug where I can't say if it's our code or some regression or incompatibility between libraries.

src/main/java/minijava/ir/NumberOfLocalVariablesVisitor.java
```diff
@@ -0,0 +1,120 @@
...
+    return 0;
+  }
+
+  @Override
+  public Integer visitIf(Statement.If that) {
+    int elseNum = 0;
+    if (that.else_.isPresent()) {
+      elseNum = that.else_.get().acceptVisitor(this);
+    }
+    return Math.max(that.then.acceptVisitor(this), elseNum);
```
sgraf812, line 43: Something that I can say after understanding Firm variables: we should not take the maximum here, but allocate a new variable index for each Localvarbinding we encounter. So this should be `+` instead of `max`.

I was under the false impression that firm variables were for stack allocation rather than for constructing SSA form, where taking the maximum would've made sense.

PR #88, opened 2016-12-03
Allow int x = (x = 1); but reject int x = x + 1;

sgraf812:
This would be prettier with some parameter passing, but, well, let's hope we don't touch `SemanticAnalyzer` too often anymore.


sgraf812:
Happy to say, we don't need this https://github.com/mj3-16/mjtest/issues/52

PR #97, opened 2016-12-16
Redirect constant control flow when matching on a Phi with constant input

sgraf812:
This optimizes away the inefficiencies introduced with modeling booleans as `b` nodes.

Example svg in Slack :)


src/main/java/minijava/ir/optimize/BaseOptimizer.java
```diff
@@ -3,8 +3,9 @@
 import firm.BackEdges;
 import firm.Graph;
 import firm.nodes.Node;
+import firm.nodes.NodeVisitor;
 
-public abstract class BaseOptimizer extends DefaultNodeVisitor implements Optimizer {
+public abstract class BaseOptimizer extends NodeVisitor.Default implements Optimizer {
```
sherter, line 8: Oh cool, didn't know this existed! ;)

src/main/java/minijava/ir/optimize/UnreachableCodeRemover.java
```diff
@@ -7,11 +7,15 @@
...
+    // find and replace unreachable code with Bad nodes
+    binding_irgopt.remove_unreachable_code(graph.ptr);
+
+    // checking whether a change on the graph occurred doesn't seem to be possible
     binding_irgopt.remove_bads(graph.ptr);
-    return false; // checking whether a change on the graph occurred doesn't seem to be possible
+
+    return false;
   }
 }
```
sherter, line 21: Was this committed by accident?

src/main/java/minijava/ir/optimize/ConstantFolder.java
```diff
@@ -185,4 +185,33 @@ public void visit(Shrs node) {
   public void visit(Sub node) {
     visitBinaryOperation((lhs, rhs) -> lhs.sub(rhs), node, node.getLeft(), node.getRight());
   }
+
+  @Override
+  public void visit(Phi node) {
+    int n = node.getPredCount();
```
sherter: I don't understand: Why does constant folding need to change when implementing control flow optimization???

src/main/java/minijava/ir/optimize/ConstantControlFlowOptimizer.java
```diff
@@ -14,7 +18,7 @@
  * <p>The {@link Proj} node that is no longer relevant is replaced with a {@link Bad} node. A
  * subsequent run of an {@link Optimizer} that removes such nodes is required.
  */
```
sherter, line 16: I hoped you would implement this in a separate class. I guess the name of this class suggested that this actually belongs here ;)
If you really don't want to put these optimizations in separate classes (I would prefer that though), then please adjust the class documentation.

src/main/java/minijava/ir/optimize/ConstantControlFlowOptimizer.java
```diff
@@ -38,15 +53,229 @@ public void visit(Cond node) {
...
+    // This can blow up quadratically otherwise, I think.
+    Node newPhi = graph.newPhi(usage, preds, preferredDef.getMode());
+    //System.out.println("    Combined definitions into " + newPhi + " of block " + usage);
+    return newPhi;
+  }
+
+  private static List<Phi> getPhiNodes(Block block) {
+    return seq(BackEdges.getOuts(block)).map(be -> be.node).ofType(Phi.class).toList();
+  }
+
```
sherter: This looks all real complicated! Generally I prefer to have ideas expressed in code and not in comments. However, as long as _you_ maintain this, I'm won't stop this from being merged... ;)

sherter (changes requested):

src/main/java/minijava/ir/optimize/UnreachableCodeRemover.java
sgraf812: probably :D I had played around with it because some stuff broke, but later on notice that wasn't broken any longer... so yeah, I'll clean that up

src/main/java/minijava/ir/optimize/ConstantControlFlowOptimizer.java
sgraf812: Yeah, I'll rather pull it out into a different class. I have absolutely no clue how to name it, though...

src/main/java/minijava/ir/optimize/ConstantControlFlowOptimizer.java
sgraf812: I tried hard to name stuff in a way that makes sense, but you can only get so far with naming things, I'm afraid... At some point you need to drop comments as to *why* the code what it does, so that's what the comments are for. 

It looks really complicated, because it actually is, much more so than I hoped it would be. Most complexity comes from the fact that we have to deal with the changed dominance frontiers due to the control flow redirects, which basically invalidates some usages.

src/main/java/minijava/ir/optimize/ConstantFolder.java
sgraf812: clarifying svg

If you look at the inputs to phi node 80 in block 69, there is this reference to Cmp 73.
We know from the incoming control flow edge 0 that that must evaluate constantly to false. This is what this does. I think I'll move it into a separate pass.

sgraf812:
OK, this should be it. Btw., when we finally merge this, we should use GitHub's 'Squash and merge' option.

sherter (approved):

PR #98, opened 2016-12-16
Remove Blocks that contain a single Jmp node and nothing else

sherter:


src/main/java/minijava/ir/optimize/JmpBlockRemover.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  private boolean isJmpBlock(Block block) {
+    List<Node> nodesInBlock = nodesInBlock(block);
+    return nodesInBlock.size() == 1 && nodesInBlock.get(0) instanceof Jmp;
+  }
+
+  private List<Node> nodesInBlock(Block block) {
+    return StreamSupport.stream(BackEdges.getOuts(block).spliterator(), false)
```
sgraf812: Looks fine so far, maybe just a recommendation: There's [Seq.seq](http://www.jooq.org/products/jOO%CE%BB/javadoc/0.9.11/org/jooq/lambda/Seq.html#seq-java.lang.Iterable-) and [Seq.of](http://www.jooq.org/products/jOO%CE%BB/javadoc/0.9.11/org/jooq/lambda/Seq.html#of-T...-) which I use most of the time to convert from `Iterable<T>` or `T[]` to a `Stream<T>`. A neat bonus is that we get many more useful functions with `Seq<T>`.

You may have noticed that already, preferring plain Java CL features, but I think repeating this stuff every time we want to turn an iterable into a stream is quite a mouthful.

parttimenerd:
The only failing test case is `store_eval_order` but has nothing to do with this PR.

PR #99, opened 2016-12-19
Draft of the compiler backend

parttimenerd:
Basic implementation of the compiler backend that uses a simple stack node allocator.
It should succeed on most test cases and avoids the lost copy problem.

Edit:
This now a pull request for the feature branch that contains the following

- implementation of the assembler backend
- modifications of CLI (according to the compiler lab specifications)
- implementation of the new language features discussed in the last compiler lab meeting in december
- implementation of `native` calls and thereby the full implementation of the Itanium ABi (see #102 for more information)


src/main/java/minijava/ir/DefaultNodeVisitor.java
```diff
@@ -0,0 +1,170 @@
+package minijava.ir;
+
+import firm.nodes.*;
+
+public interface DefaultNodeVisitor extends NodeVisitor {
```
sgraf812, line 5: Hmm... I'm not happy with this. Check out `NodeVisitor.Default`.  Can't you use that instead of reimplementing all this? I find it much more comfortable to have that `defaultVisit` method which can be overriden.

src/main/java/minijava/ir/assembler/NodeAllocator.java
```diff
@@ -0,0 +1,46 @@
...
+  Location getResultLocation(Node node);
+
+  /** Returns the size of the activation record in bytes. */
+  int getActivationRecordSize();
+
+  Argument getAsArgument(Node node);
+
+  String getActivationRecordInfo();
+
+  Location createNewTemporaryVariable();
```
sgraf812: Soo... are these last four methods per graph or per block? I'm a little uncomfortable about the assumptions this makes

src/main/java/minijava/Cli.java
```diff
@@ -32,7 +36,7 @@
       Joiner.on(System.lineSeparator())
           .join(
               new String[] {
-                "Usage: minijavac [--echo|--lextest|--parsetest|--check|--compile] [--help] file",
+                "Usage: minijavac [--echo|--lex-test|--parse-test|--check|--compile-firm|--print-asm] [--help] file",
```
sgraf812: What's the reason for the added hyphens?

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
```diff
@@ -0,0 +1,584 @@
...
+    this.info = new MethodInformation(graph);
+    this.allocator = allocator;
+    allocator.process(graph);
+    blocksToCodeBlocks = new HashMap<>();
+    segment = new CodeSegment(new ArrayList<>(), new ArrayList<>());
+    segment.addComment(String.format("Code segment for method %s", info.name));
+    blocksToCodeBlocks = new HashMap<>();
+  }
+
+  public Segment generate() {
```
sgraf812: Generate... what exactly? Judging from the implementation this generates Code for the passed graph.

Hmm... Why not make the graph a parameter of the method? That would be a lot clearer. I really don't like (unnecessary?) instance state like that.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
```diff
@@ -0,0 +1,584 @@
...
+    BackEdges.enable(graph);
+    graph.walkTopological(this);
+    prependStartBlockWithPrologue();
+    segment.addComment(allocator.getActivationRecordInfo());
+    return segment;
+  }
+
+  public static AssemblerFile generateForAll(Supplier<NodeAllocator> nodeAllocCreator) {
```
sgraf812: `generateForAll` ... for all what? Maybe pass the list of graphs here explicitly instead of relying on firm global state.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
```diff
@@ -0,0 +1,584 @@
...
+              .com(
+                  "Store the first argument in an intermediate register, as "
+                      + "the assembler uses a two adress format (the first argument "
+                      + "has to be the result location)"));
+      block.add(creator.apply(intermediateReg, secondArg).firm(node));
+      block.add(new Mov(intermediateReg, res));
+    } else {
+      // everything is fine here
+      block.add(creator.apply(firstArg, secondArg).firm(node));
+    }*/
```
sgraf812: Soo.. What happened to this hack?

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
```diff
@@ -0,0 +1,584 @@
...
+    Argument dest = allocator.getAsArgument(node.getPtr());
+    Argument newValue = allocator.getAsArgument(node.getValue());
+    // copy the new value into a register
+    block.add(new Mov(newValue, intermediateValReg).firm(node.getValue()));
+    // copy the address into another register
+    block.add(new Mov(dest, intermediateDestReg).firm(node.getPtr()));
+    // store the new value at its new location
+    block.add(new Mov(intermediateValReg, new StackLocation(intermediateDestReg, 0)).firm(node));
+  }
+}
```
sgraf812, line 619: Hmm... I'm generally not happy with the fact that we just assume EAX and EBX to always be available for loading arguments. This should reside in register allocation, but I'm not sure how we can model that, so we'll have to live with that for now, I think.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
```diff
@@ -0,0 +1,584 @@
...
+    Register intermediateValReg = Register.EAX;
+    Register intermediateDestReg = Register.EBX;
+    Argument dest = allocator.getAsArgument(node.getPtr());
+    Argument newValue = allocator.getAsArgument(node.getValue());
+    // copy the new value into a register
+    block.add(new Mov(newValue, intermediateValReg).firm(node.getValue()));
+    // copy the address into another register
+    block.add(new Mov(dest, intermediateDestReg).firm(node.getPtr()));
+    // store the new value at its new location
+    block.add(new Mov(intermediateValReg, new StackLocation(intermediateDestReg, 0)).firm(node));
```
sgraf812: `StackLocation` seems like kind of a misnomer if we also use it for memory access. Maybe rename it to `MemoryLocation` (maybe even without Location)?

src/main/java/minijava/ir/assembler/SimpleNodeAllocator.java
```diff
@@ -0,0 +1,171 @@
...
+  private int currentSlotNumber = 1;
+  private Graph graph;
+  private MethodInformation info;
+
+  @Override
+  public void process(Graph graph) {
+    this.graph = graph;
+    this.info = new MethodInformation(graph);
+    this.assignedStackSlots = new HashMap<>();
+  }
```
sgraf812, line 49: Ah, now the purpose of the interface method is more clear.

Do we really need this `process` method? This just seems to initialize stuff and hide the graph instance, which I'd rather see passed as an explicit argument to methods which need it. As I said, I really dislike instance state which leaks out of an interface.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
```diff
@@ -0,0 +1,126 @@
+package minijava.ir.assembler.block;
+
+import com.sun.jna.Platform;
+import java.util.*;
+import minijava.ir.NameMangler;
+import minijava.ir.assembler.GNUAssemblerConvertible;
+import org.jetbrains.annotations.NotNull;
+
+/** List of segments that are combined to an assembler file */
+public class AssemblerFile implements GNUAssemblerConvertible, Collection<Segment> {
```
sgraf812, line 13: Hmm... implementing the whole `Collection<>` interface seems quite unnecessary when all we need is a way to `addSegment`s to an `AssemblerFile`. YAGNI. I'd also go so far and say that we only need a static method `GNUAssembler.dumpCode(segments) -> String` and save that String in a file.

src/main/java/minijava/ir/assembler/block/CodeBlock.java
```diff
@@ -0,0 +1,224 @@
...
+import minijava.ir.assembler.GNUAssemblerConvertible;
+import minijava.ir.assembler.instructions.Cmp;
+import minijava.ir.assembler.instructions.Instruction;
+import minijava.ir.assembler.instructions.Jmp;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A list of assembler instructions with a label, optional cmp support, cmp and jmp instructions.
+ */
+public class CodeBlock implements GNUAssemblerConvertible, Collection<Instruction> {
```
sgraf812: Again: Implementing the whole `Collection<>` is more misleading than anything, especially if all you ever do is `add` or check `isEmpty`. Especially since you throw `UnsupportedOperationException` most of the time.

If you *really* want to have the collection interface, expose the `normalInstructions` field in a getter.

Generally implementing `Collection`/`List`/`Iterable` etc. interfaces is a code smell. Most of the time you can the relevant logic into a collection class with a single responsibility (as opposed to these frankensteins) or even just expose a field.

src/main/java/minijava/ir/assembler/block/CodeSegment.java
```diff
@@ -0,0 +1,47 @@
...
+    blocks.add(block);
+  }
+
+  public void addComment(String comment) {
+    comments.add(comment);
+  }
+
+  public List<CodeBlock> getBlocks() {
+    return Collections.unmodifiableList(blocks);
+  }
```
sgraf812, line 46: See, this is much better :)

src/main/java/minijava/ir/utils/FirmUtils.java
```diff
@@ -0,0 +1,46 @@
...
+  public static String getMethodLdName(firm.nodes.Call node) {
+    return ((Address) node.getPred(1)).getEntity().getLdName();
+  }
+
+  public static boolean isPhiProneToLostCopies(Phi phi) {
+    if (!isPhiProneToLostCopies.containsKey(phi)) {
+      isPhiProneToLostCopies.put(phi, checkPhi(phi));
+    }
+    return isPhiProneToLostCopies.get(phi);
+  }
```
sgraf812: Please, don't do this. No static instance state in a helper class :( Just use `checkPhi` each time. This seems like premature optimization.

src/main/resources/mj_runtime.c
```diff
@@ -17,6 +17,12 @@ int main() {
 }
 
 void* calloc_impl(size_t num, size_t size){
+  //printf("num=%zu size=%zu ", num, size);
+  // the upper 4 bytes may contain rubish
+  // and we don't need to allocate such large memory areas at once
+  // we can still allocate around 4 Gbytes
+  num = num & 0x00000000ffffffff;
+  size = size & 0x00000000ffffffff;
```
sgraf812, line 37: hmm... maybe we can replace the `size_t`  parameters by `int32_t` (again)? That should make the C compiler do the necessary extensions, I think.

sgraf812 (commented):
Wow, you were quite busy :+1: I haven't tried yet, but the tests are green, so...

src/main/java/minijava/ir/DefaultNodeVisitor.java
parttimenerd: I like to have a NodeVisitor interface. I'm not happy with using an abstract class.

src/main/java/minijava/ir/assembler/NodeAllocator.java
parttimenerd: Yes. The assumption is debatable, but it can be changed later.

src/main/java/minijava/Cli.java
parttimenerd: Just a mistake.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
parttimenerd: You're correct, I change it soon.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
parttimenerd: Method naming is hard.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
parttimenerd: Correct. I rename it to `MemoryLocation`.

src/main/java/minijava/ir/assembler/SimpleNodeAllocator.java
parttimenerd: The method initializes the allocator for a given graph.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
parttimenerd: I don't think so. I like to encapsulate things in objects...

src/main/java/minijava/ir/assembler/block/CodeBlock.java
parttimenerd: Taken.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
sgraf812: Yeah, but what are you actually encapsulating here? Looking at the above function signature makes it pretty clear what it's doing, while having some class with some tenish instance methods, most of them modifying some hidden state isn't clear at all.

src/main/java/minijava/ir/utils/FirmUtils.java
parttimenerd: Why? `checkPhi` is costly as we traverse the whole graph and the methods is used quite often (several times on each phi node).

src/main/resources/mj_runtime.c
parttimenerd: Good idea.

src/main/java/minijava/ir/utils/FirmUtils.java
sgraf812: Then maybe cache it on the use site? I still don't think that performance would be much of a problem if we didn't.

parttimenerd: I don't see the problem of caching it here either.

sgraf812: Well, for one it assumes that the predecessors of the cached `Phi`s aren't changed between invocations of `isPhiProneToLostCopies`. Meaning that if at one point, we decide to implement an optimization which might be able to change predecessors of a `Phi` after it is sure that the `Phi` `isPhiProneToLostCopies`, the results might change. In other words: You memoize a function which is not pure, producing different output depending on where/when you call it.

I'd move the caching component completely into the `AssemblyGenerator`, which is the only place where you can maybe live with that invariant. `checkPhi` may stay in this public helper, but the caching wrapper is pretty much unsafe to use outside of the `AssemblyGenerator`.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
parttimenerd: It was discarded after assumption about registers we're made.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
parttimenerd: This class could for example join adjacent segments of the same type.

src/main/java/minijava/ir/utils/FirmUtils.java
parttimenerd: Taken.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
sgraf812: A static method could do that equally well, I'd say. But whatever, this isn't really world changing.

src/main/resources/mj_runtime.c
parttimenerd: But we could also improve the generated assembler.

src/main/java/minijava/ir/assembler/AssemblerGenerator.java
parttimenerd: After some considerations: i'll rename the method, but inializing stuff in the constructor doesn't seem to be wrong to me. That's the purpose of constructors after all.

src/main/java/minijava/ir/DefaultNodeVisitor.java
sherter: A NodeVisitor interface is provided by jFirm. Its name is: `NodeVisitor`! ;)
What you want to do here is providing default _behavior_ and that's not something an interface should be used for (yes, Java 8 _can_ do that, but I don't think it was intended to be used like that).

parttimenerd: Using `NodeVisitor` litters the class with lot's of methods with empty bodies.

sgraf812: Yeah, that's why it's good to have `NodeVisitor.Default`. I can see why you want an interface there (mixin inheritance >>> abstract classes), but it hardly justifies not using what's already there and used on master in the meantime. Also, in the cases where you use `DefaultNodeVisitor`, you *could* use an abstract class instead, e.g. there are no other base classes.

parttimenerd: You're correct. Maybe just stubborness.

parttimenerd: i removed any usage of this class.

PR #100, opened 2016-12-28
Support cli calls without options and make graph output optional

parttimenerd:


sgraf812:
Although I dislike all these environment variables, I understand why they are necessary... Well, both PRs look good to me, but remember to resolve the merge conflicts on #99 at some point so that we can merge it.

PR #104, opened 2017-01-04
White list for google java format

sgraf812:
White listing `src` now instead of blacklisting some of the directories. The check is quite annoying when having temporary top-level java files for poking around in the implementation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ repositories {
 
 googleJavaFormat {
     toolVersion '1.1'
-    include 'src'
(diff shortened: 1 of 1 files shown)


sherter:
Instead of first adding everything and then filter files with path `src` (that's what include does) it would be better to scan only the `src` directory and filter for `*.java` files.  Or even better, use
```
googleJavaFormat {
  source = sourceSets*.allJava
}
```
which already includes the filter for `*.java` files.

PR #106, opened 2017-01-04
Remove phi of phi

sgraf812:
E.g. if there is code like `Phi(..., Phi(..., a))` we can optimize that to `Phi(..., a)`, if

- The outer phi is not the inner phi (can happen in loops)
- `a` is not already the inner phi (in which case there's nothing to do)
- Both phis are in the same block


sgraf812:
OK, turns out this wasn't a good idea, as it's not a valid optimization.

PR #107, opened 2017-01-04
Implemented Common Subexpression Elimination

sgraf812:


sgraf812:
This fails because currently I don't respect dominance relations, e.g. if the exchanged subexpression is actually visible at the use site. I already have a solution in mind, but that's not gonna happen tonight.

sgraf812:
Doing CSE might not always be beneficial, I think... Cheap expressions might not be worth to share, because they raise register pressure. But generally we should run with this.

sherter:
You could use http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash(java.lang.Object...) for hash calculation, otherwise :+1:

sgraf812:
Thanks for the pointer, will do.

sgraf812:
Something breaks CI, I'm not sure what. This is on hold until I figure it out.

Edit: Ah, it's just the old backend. So let's merge #110 first.

PR #108, opened 2017-01-06
Normalize expressions

sgraf812:
This orders arguments to `Add`, `Mul` and `Cmp` nodes in a (hopefully) well-defined order, defined by `ExtensionalEqualityComparator`.

### DIFF
--- a/src/main/java/minijava/Cli.java
+++ b/src/main/java/minijava/Cli.java
@@ -180,6 +180,7 @@ public class Cli {
     Optimizer constantFolder = new ConstantFolder();
     Optimizer controlFlowOptimizer = new ConstantControlFlowOptimizer();
     Optimizer unreachableCodeRemover = new UnreachableCodeRemover();
+    Optimizer expressionNormalizer = new ExpressionNormalizer();
     Optimizer algebraicSimplifier = new AlgebraicSimplifier();
     Optimizer commonSubexpressionElimination = new CommonSubexpressionElimination();
     Optimizer phiOptimizer = new PhiOptimizer();
@@ -187,6 +188,7 @@ public class Cli {
     for (Graph graph : firm.Program.getGraphs()) {
       //Dump.dumpGraph(graph, "before-simplification");
       while (constantFolder.optimize(graph)
+          | expressionNormalizer.optimize(graph)
           | algebraicSimplifier.optimize(graph)
           | commonSubexpressionElimination.optimize(graph)
           | phiOptimizer.optimize(graph)) ;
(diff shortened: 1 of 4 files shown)


parttimenerd:
Could you reorder the node predecessors in a way that constant arguments are the first arguments, except for `Sub` where they should be the last argument?

sgraf812:
Hmm... That would be an annoying special case, but doable (add a special case checking for `Const` [here](https://github.com/mj3-16/minijavac/pull/108/files#diff-06a93850e7402f492471e755424e13a6R47)).

Also I don't think we should put the `Sub` to `Add + Neg` transformation in the middle-end, seems to be more of an architecture specific backend issue.

parttimenerd:
Placing it in the middle end has the advantage of other optimizations possibly optimizing the `Neg` away.
But yes, I agree that this is a special case.

sgraf812:
Any reason why you want `Const`s to be the leftmost arguments? The algebraic simplifier states simplification rules with consts as right arguments... Doesn't really make a huge difference, but I'd prefer consts to be right arguments, makes more sense for e.g. pointer arithmetic (where I'm used to have base + const offset).

parttimenerd:
Th reason is the used GCC assembler. For `Add` and `Mul` the right argument is the destination. And for `Sub` it's the other way around as I have to swap the arguments for assembler generation (`x - y` equals `sub y, x`).
It greatly simplifies the assembler generation.

sgraf812:
Hmm... I'm really not sold on leaking backend stuff into the IR. Couldn't you just work with the assumption that `Const`s are always right arguments? You probably have to check that anyway, because the backend shouldn't assume that constants are folded, e.g. we should still be able to generate code for `Add(Const, Const)`.

parttimenerd:
The new backend makes the explicit assumption that the constant folding and normalizing step run before. The reason for this are my bad experiences while implementing the old backend. Not making these assumptions made the code really cluttered with many special cases.

sgraf812:
We should then have a separate pass that does the required transformations just before codegen, which in this case would only be the `Sub` to `Add + Neg` thing. The last commit should be enough to order `Add`/`Mul` arguments in a way that `Const`s are first.

parttimenerd:
Okay. When I merge the new backend, I'll modify me Normalizer. Are you okay with the assumption that the constant folder is used before running the backend?

sgraf812:
OK for now, I guess... I'm sure we'll work something out eventually.

sherter:
Generally, I agree that the backend should be able to handle what ever (valid) IR we give it. However, as long as the assumptions you make about the IR are documented somewhere, I'm okay with it ;)

sgraf812:
We talked about this and came to the conclusion that his `Normalizer` did somewhat different things than this `Normalizer`, e.g. his ones lines up `Const` nodes so that the code generator won't freak out, wheras this is simply an enabling transformation for stuff like CSE. I think it's separate for good reason and I don't see the point in combining two unrelated transformations.

parttimenerd:
The current (simplified) register allocator doesn't make any assumptions in this area. Therefore we don't need a backend specific normalizer.

PR #110, opened 2017-01-11
Reimplement the backend

parttimenerd:
Reimplements that backend and allows to create more advanced register allocators.


src/main/java/minijava/ir/assembler/SimpleNodeAllocator.java
```diff
@@ -70,25 +64,28 @@ public Location getLocation(Node node) {
...
     }
-    return getStackSlotAsLocation(node);
+    if (node instanceof firm.nodes.Call) {
+      return getNodeLocation(
+          node,
+          ((MethodType) (((Address) node.getPred(1)).getEntity().getType()))
+              .getResType(0)
+              .getMode());
+    }
+    if (node instanceof firm.nodes.Load) {
```
sherter, line 85: The `instanceof` checks seem wrong here, why not use a visitor?

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
```diff
@@ -127,11 +130,26 @@ private static String getGNUAssemblerFilePrologue() {
...
+      CodeSegment segment = asmGenerator.generateSegmentForGraph();
+      preAsmFile.add(segment);
+      LinearCodeSegment linCode = LinearCodeSegment.fromCodeSegment(segment);
+      try {
+        BasicRegAllocator allocator =
+            new BasicRegAllocator(
+                new MethodInformation(graph), linCode, asmGenerator.getAllocator());
+        LinearCodeSegment ret = allocator.process();
+        file.add(ret);
+      } catch (NullPointerException ex) {
```
sherter, line 148: That stinks quite a bit...

src/main/java/minijava/ir/assembler/block/LinearCodeSegment.java
```diff
@@ -0,0 +1,118 @@
...
+      List<Instruction> instrs = new ArrayList<>();
+      instrs.addAll(newInstructions);
+      instrs.add(instruction);
+      replaceInstruction(instruction, instrs);
+    }
+
+    public void appendToEnd(List<Instruction> instructions) {
+      placeAfter(instructions.get(instructions.size() - 1), instructions);
+    }
+  */
```
sherter, line 114: If we don't need it, remove it! (it's still in version control in case we actually need it again)

sherter (commented):
It really doesn't make much sense to review the changes here, there are simply too many!
I left a view comments, but I guess we should merge it and maybe review/improve it incrementally afterwards...

src/main/java/minijava/ir/assembler/SimpleNodeAllocator.java
sgraf812: Visitor is most useful for making sure not to have forgotten a case at compile time. If you aren't matching all cases anyway, using a visitor just adds noise IMO.

Edit: Also I think it's a shame that jFirms API relies on having to pass a `NodeVisitor` instead of just a `Consumer<Node>`.  You can always turn a `NodeVisitor` into a `Consumer<Node>` by `n -> n.accept(visitor)`, it's not so easy the other way around. That's why there are ugly hacks such as `NodeVisitor.Default`. But whatever, we have to live with what we got.

src/main/java/minijava/ir/assembler/block/AssemblerFile.java
parttimenerd: No. It just helps with debugging.

PR #112, opened 2017-01-12
Implement an inliner

sgraf812:
This is most useful as an enabling transformation and should bring a big difference in interplay with the other transformations we do. Also the [heuristic](https://github.com/mj3-16/minijavac/blob/f9d90aec0d4a210709db3988f49b58bc5d70d48d/src/main/java/minijava/ir/optimize/Inliner.java#L224-L238) of what to inline is rather crude: We only inline functions which don't call functions themselves, that don't diverge, and that don't contain more than 1000 nodes.


sherter:
I might have split up some of the bigger methods even more, but the code is quite comprehensible. Well done! :wink:

PR #117, opened 2017-01-15
Implement an on-the-fly register allocator

parttimenerd:
It reduces the runtime of programs in some benchmarks by one third.

### DIFF
--- a/src/main/java/minijava/Cli.java
+++ b/src/main/java/minijava/Cli.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Optional;
 import minijava.ast.Program;
 import minijava.ir.IREmitter;
+import minijava.ir.assembler.allocator.OnTheFlyRegAllocator;
 import minijava.ir.assembler.block.AssemblerFile;
 import minijava.ir.optimize.*;
 import minijava.ir.utils.CompileUtils;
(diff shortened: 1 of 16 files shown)


sherter:
I don't quite understand why you added the previously removed code back in? Most of it is still never used?!

parttimenerd:
Can you please give some examples for your claim? Some methods might not be used, but most added methods should be used.

sherter:
Maybe we should add an external tool that provides such information, if your IDE can't do that for you.
Examples in `CodeBlock` are:
- size()
- getAfterCompareInstructions()
- addAll()
- hasUnconditionalJump()
- getCompare() 
- getUnconditionalJump()
- getConditionalJumps()
- getBlockStartInstructions()

parttimenerd:
My IDE can do that for me. I forgot to say that I remove some unused methods soon.
But I dislike the notion of removing all methods that aren't used currently. Especially if there's a non zero chance that they'll be used in the future (e.g. I use the `size` method in a new optimization I'm currently developing).

sgraf812:
Well, I'd rather see them removed. Even if I'd need similar functionality in the future, I probably won't find it and just unintentionally duplicate stuff.

parttimenerd:
Good point. I'll remove them soon.

parttimenerd:
@sherter is code acceptable for you (I removed all the unused methods)?

sherter:
The field `methodInfo`  in `AbstractRegAllocator` is also unused and `isUsedAfterInstruction(Instruction instruction)` in `Argument.InstructionRelations`. There are a few others but they might not be related to this PR...

parttimenerd:
I use the `isUsedAfter` method in the optimization I'm currently developing.

sherter:
okay then, ship it ;)  :+1:

PR #125, opened 2017-01-17
Executing simplifier again after control flow changes

sgraf812:
... So that optimization opportunities after control flow optimizations can be observed by the simplifier.

It's probably high time that the optimization order gets abstracted into a seperate, higher-order fixed-point iteration.


sgraf812:
Hmm... We probably should loosen the time limit in mjtest.

PR #132, opened 2017-01-18
Model per graph optimizations as higher-order fixed-point iteration

sgraf812:
This is much more declarative and it's immediately clear when an optimization fires, since they state their dependencies.

### DIFF
--- a/src/main/java/minijava/ir/optimize/JmpBlockRemover.java
+++ b/src/main/java/minijava/ir/optimize/JmpBlockRemover.java
@@ -5,7 +5,6 @@ import static org.jooq.lambda.Seq.seq;
 import com.google.common.collect.Iterables;
 import firm.BackEdges;
 import firm.Graph;
-import firm.Mode;
 import firm.nodes.Block;
 import firm.nodes.Jmp;
 import firm.nodes.Node;
@@ -20,14 +19,14 @@ public class JmpBlockRemover extends BaseOptimizer {
     this.graph = graph;
     this.hasChanged = false;
     BackEdges.enable(graph);
-    graph.walkTopological(this);
+    graph.walkPostorder(this);
     BackEdges.disable(graph);
     return hasChanged;
   }
 
   @Override
   public void visit(Block block) {
-    if (!hasChanged && isJmpBlock(block) && hasOnePredecessor(block)) {
+    if (isJmpBlock(block) && hasOnePredecessor(block)) {
       hasChanged = true;
       remove(block);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/minijava/ir/optimize/JmpBlockRemover.java
```diff
@@ -57,10 +56,6 @@ private void remove(Block block) {
     jmpTargetNewPredecessors[jmpToTargetEdge.pos] = jmpBlockPredecessor;
     Node newJmpTarget = graph.newBlock(jmpTargetNewPredecessors);
     Graph.exchange(jmpTarget, newJmpTarget);
-
-    Node bad = graph.newBad(Mode.getANY());
-    for (int i = 0; i < block.getPredCount(); i++) {
-      block.setPred(i, bad);
-    }
+    Graph.killNode(block);
```
sherter, line 59: Oh, nice! ;)

src/main/java/minijava/ir/optimize/JmpBlockRemover.java
```diff
@@ -20,14 +19,14 @@ public boolean optimize(Graph graph) {
...
-    graph.walkTopological(this);
+    graph.walkPostorder(this);
     BackEdges.disable(graph);
     return hasChanged;
   }
 
   @Override
   public void visit(Block block) {
-    if (!hasChanged && isJmpBlock(block) && hasOnePredecessor(block)) {
+    if (isJmpBlock(block) && hasOnePredecessor(block)) {
```
sherter, line 29: This is possible now because `walkPostorder` doesn't break if the graph is changed while traversing it, right?

sherter (approved):
I like it!

src/main/java/minijava/ir/optimize/JmpBlockRemover.java
sgraf812: Yep, at least I hope so. Back edges seem problematic, but I've not had any problems so far.

src/main/java/minijava/ir/optimize/JmpBlockRemover.java
sgraf812: Yeah, sorry for sneaking these refactorings in. But I felt it wasn't just worth the time to open a separate PR.

src/main/java/minijava/ir/optimize/JmpBlockRemover.java
sgraf812: Also I don't think that we need the topological order at all, since we don't actually change any control flow paths. Well, it's mostly by stomach telling me that it's safe.

I changed it in the first place because `graph.walk*` variants would not visit all nodes, which is really strange and might have to do with the interplay of some global state in firm, who knows...

PR #135, opened 2017-01-18
Inline until reaching a certain maximum of nodes in the graph

sherter:
We can always play with the actual MAX_SIZE. I'm interested in how this affects the runtime performance in speedcenter ;)

### DIFF
--- a/src/main/java/minijava/ir/optimize/Inliner.java
+++ b/src/main/java/minijava/ir/optimize/Inliner.java
@@ -29,13 +29,13 @@ import minijava.ir.utils.GraphUtils;
 import minijava.ir.utils.NodeUtils;
 import org.jooq.lambda.tuple.Tuple2;
 
-public class Inliner extends BaseOptimizer {
+abstract class Inliner extends BaseOptimizer {
   private static final Set<ir_opcode> ALWAYS_IN_START_BLOCK =
       Sets.newHashSet(iro_Const, iro_Address);
-  private final ProgramMetrics metrics;
-  private final Set<Call> callsToInline = new HashSet<>();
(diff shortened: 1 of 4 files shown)


sgraf812:
Hmm... I don't think we actually need different classes for this. The difference should more or less just be a flag. Stuff like the size limit is also useful for the `LeafInliner`, I think.

Edit: Also, as always, I dislike abstract classes. Why not extract an interface `InlineAbilityDecider` instead?

sherter:
That was actually the approach I used at first but then discovered that I would need to modify the inlineCandidates() method, too. I couldn't find a way to do that cleanly, so I decided to just use inheritance... I know, you are not a fan of it (me neither), but it seemed to be the simplest solution.

sherter:
I couldn't find a way to do that cleanly, so I decided to just use inheritance... I know, you are not a fan of it (me neither), but it seemed to be the simplest solution.

sgraf812:
Because of the size limit? I think the `LeafInliner` should obey that limit, too. I was just too lazy to implement it in a correct way (like you did now)

sherter:
I thought it would make sense to always inline leaf methods (with less than 1000 nodes), even if the overall size is bigger than the MAX_SIZE limit. Just a feeling, though....

sgraf812:
That's really debatable, but I think the same arguments apply as for recursive functions: The call overhead is negligible in comparison.

Although if that would mean we can't inline a getter within a loop, that would be bad...

Anyway, I'd much rather see this implemented as a boolean flag. There is so much duplication of important logic (mainly the inlining heuristic) that I think we could tolerate the little ugliness that would bring.

PR #136, opened 2017-01-19
Also inline non-leaf methods

sherter:
Another approach for inlining non-leaf methods. This time based on a flag, as requested ;)

### DIFF
--- a/src/main/java/minijava/ir/optimize/Inliner.java
+++ b/src/main/java/minijava/ir/optimize/Inliner.java
@@ -64,8 +64,7 @@ public class Inliner extends BaseOptimizer {
     for (Call call : callsToInline) {
       int methodSize = metrics.graphInfos.get(call.getGraph()).size;
       int newGraphSize = methodSize + size;
-      if (onlyLeafs && (methodSize < MAX_LEAF_SIZE_TO_ALWAYS_INLINE || newGraphSize < MAX_NODES)
-          || (!onlyLeafs && newGraphSize < MAX_NODES)) {
+      if (onlyLeafs && methodSize < MAX_LEAF_SIZE_TO_ALWAYS_INLINE || newGraphSize < MAX_NODES) {
         inline(call);
         hasChanged = true;
         size = newGraphSize;
--- a/src/main/java/minijava/ir/optimize/Inliner.java
+++ b/src/main/java/minijava/ir/optimize/Inliner.java
@@ -32,27 +32,44 @@ import org.jooq.lambda.tuple.Tuple2;
 public class Inliner extends BaseOptimizer {
   private static final Set<ir_opcode> ALWAYS_IN_START_BLOCK =
       Sets.newHashSet(iro_Const, iro_Address);
+  private static final int MAX_NODES = 1000;
+  /**
+   * Specifies the maximal number of nodes a leaf method can have to always be inlined, regardless
+   * of whether the graph already has more than {@link #MAX_NODES} nodes or not.
+   */
+  private static final int MAX_LEAF_SIZE_TO_ALWAYS_INLINE = 80;
+
   private final ProgramMetrics metrics;
   private final Set<Call> callsToInline = new HashSet<>();
 
-  public Inliner(ProgramMetrics metrics) {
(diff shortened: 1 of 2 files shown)


src/main/java/minijava/ir/optimize/Inliner.java
```diff
@@ -32,27 +32,44 @@
...
     boolean hasChanged = false;
+    int size = metrics.graphInfos.get(graph).size;
     for (Call call : callsToInline) {
-      //System.out.println("Inlining " + callee(call) + " into " + graph);
-      inline(call);
-      hasChanged = true;
+      int methodSize = metrics.graphInfos.get(call.getGraph()).size;
+      int newGraphSize = methodSize + size;
+      if (onlyLeafs && (methodSize < MAX_LEAF_SIZE_TO_ALWAYS_INLINE || newGraphSize < MAX_NODES)
+          || (!onlyLeafs && newGraphSize < MAX_NODES)) {
```
sgraf812: How about `onlyLeafs && methodSize < MAX_LEAF_SIZE_TO_ALWAYS_INLINE || newGraphSize < MAX_NODES`?

I have to truth table handy, but should be the same:

```
(a & (b | c)) | (!a & b)
= distribute a &
a & b | a& c | !a & b
= law of the excluded middle on a and !a
a& c | b
```

Otherwise, we are good!

src/main/java/minijava/ir/optimize/Optimizer.java
```diff
@@ -52,23 +52,26 @@ static void optimize() {
...
       for (Graph graph : firm.Program.getGraphs()) {
         hasChanged |= inliner.optimize(graph);
         unreachableCodeRemover.optimize(graph);
-        metrics.updateGraphInfo(graph);
         Cli.dumpGraphIfNeeded(graph, "after-inlining");
       }
       if (!hasChanged) {
-        break;
+        if (inliner.onlyLeafs) {
+          inliner = new Inliner(metrics, false);
```
sgraf812, line 71: Well, one additional thing: Maybe define an `enum InliningAggressiveness { ONLY_LEAFS; EVERYTHING }`  which better describes what's going on.

Won't make much a difference, though.

sherter: Now I'm a little confused. Didn't you say you wanted a flag?
This is really close to simulating object orientation with enums. Well, the flag is basically doing the same thing, but if you _do_ want extensibility we should, in my opinion, go with real object orientation instead of enums.

src/main/java/minijava/ir/optimize/Inliner.java
sherter: Got catch! :1st_place_medal:

src/main/java/minijava/ir/optimize/Optimizer.java
sgraf812: I didn't say I didn't want object-orientation, I said that I really didn't want to have abstract classes and that instead we should extract an interface.

In this case the interface we'd extract would have exactly one method: `boolean inlineOnlyLeafs();`, where its two implementations would return `false` and `true` respectively.

You are right in that this is *exactly* like an `enum`, so that's what I suggested, but an enum captures the pattern more tersely (like the connection between interfaces + concrete classes and visitors we had when implementing AST nodes).

The thing here is that we should favor [Composition over Inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). I've yet to see a case where it's not viable to extract an interface instead of sharing logic through an abstract base class.

This might sound theoretical given the simplicity of the current problem, but what best proves that it's the right thing to do is the difference in duplicated code and overall code needed to achieve the same as the old PR.

I'm absolutely fine with the PR as it is now, just trying to explain my thinking here.

sherter: There are certainly many ways to do this in an object oriented fashion.
Anyhow - as you said it won't make much of a difference, do you want me to change it or not?

sgraf812: Ah, keep it, just wanted to kick off a discussion 💃

PR #138, opened 2017-01-20
Capture side effects of Div and Mod and ripple effects

sgraf812:
I'm sorry that this is yet again an amalgamation of multiple possible PRs:

I finally came around to fix side effects for `Div` and `Mod`. Consider this program:

```
class div_by_zero_side_effect {
        public static void main(String[] args) {
                int j = 5 / 0;
                System.out.println(0);
        }
}
```

On current master, this will print 0 and exit successfully, because we didn't `setCurrentMem` when emitting the `Div`. Since `j` is dead, the `Div` is not reachable at all from the `End` node and thus there will be no code emitted, although we certainly need to preserve semantics and throw an exception here! That's why we need the memory edge.

So I did the trivial change to `IREmitter`, but that broke `ConstantFolder`. After having understood the problem (we'd have to redirect memory edges when exchanging nodes), I noticed strange behavior: Mode M nodes were folded. And so I chased the rabbit down the hole... Of particular note is the `defaultVisit` override, which assumes `Bad` values for all unhandled nodes. Previously those had `Unknown` as values, rendering the Analysis unsound. A fortunate side-effect is that I got rid of the two different `PhiStrategy`s this way.

I again noticed, that `graph.walk*` just doesn't work and also that the visit order wasn't what we wanted for `JmpBlockRemover`, so I replaced it with a call to `GraphUtils.walkReversePostOrder`.

The changes to `PhiOptimizer` are rather modest and were the original motivation for the PR.

@parttimenerd Is it somehow possible to write tests in `mjtest` that test for errors? Otherwise I don't see how we could reliably test if we preserve error behavior...

### DIFF
--- a/src/main/java/minijava/ir/optimize/ConstantFolder.java
+++ b/src/main/java/minijava/ir/optimize/ConstantFolder.java
@@ -2,6 +2,7 @@ package minijava.ir.optimize;
 
 import static org.jooq.lambda.Seq.seq;
 
+import com.google.common.collect.Sets;
 import firm.BackEdges;
 import firm.Graph;
 import firm.Mode;
@@ -11,16 +12,17 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import minijava.ir.utils.FirmUtils;
 import minijava.ir.utils.GraphUtils;
 
-public class ConstantFolder extends NodeVisitor.Default implements Optimizer {
+public class ConstantFolder extends BaseOptimizer {
 
+  private static final Set<Mode> HANDLED_MODES =
+      Sets.newHashSet(Mode.getBu(), Mode.getb(), Mode.getIs(), Mode.getLs());
   private Map<Node, TargetValue> latticeMap;
-  private Graph graph;
-  private boolean hasChanged;
 
   @Override
   public boolean optimize(Graph graph) {
@@ -31,15 +33,7 @@ public class ConstantFolder extends NodeVisitor.Default implements Optimizer {
         () -> {
           do {
             hasChanged = false;
-            GraphUtils.walkPostOrder(
-                graph,
-                n -> {
-                  if (n.getMode().equals(Mode.getM())) {
-                    // Constant folding does not make sense for side-effects
-                    return;
-                  }
-                  n.accept(this);
-                });
+            GraphUtils.walkPostOrder(graph, this::visit);
           } while (hasChanged);
           return replaceConstants();
         });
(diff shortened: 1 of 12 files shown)


parttimenerd:
Regarding `mjtest`: It's currently not possible, but adding an appropriate test case handler is easy.
I'm going to implement it. (i didn't implement one as the compiler lab test suite only tests valid MiniJava files and invalid are deemed unimportant)

sherter:
Why are you replacing
```java
extends BaseOptimizer {
```
 everywhere with
```java
extends NodeVisitor.Default implements Optimizer {
  private Graph graph;
  private boolean hasChanged;
```
? I know you don't like inheritance, but i find duplicating code to be worse.

Just as a reminder, this is what `BaseOptimizer` looks like:
```java
public abstract class BaseOptimizer extends NodeVisitor.Default implements Optimizer {

  protected Graph graph;
  protected boolean hasChanged;
}
```

sgraf812:
Yeah, you're right, I'll change it back. Although I find it more difficult to reason about non-private fields because who knows what the base class might do to them.

src/main/java/minijava/ir/optimize/ConstantFolder.java
```diff
@@ -1,43 +1,49 @@
...
 import java.util.function.BiFunction;
 import java.util.function.Function;
-import java.util.stream.StreamSupport;
+import minijava.ir.utils.FirmUtils;
+import minijava.ir.utils.GraphUtils;
 
 public class ConstantFolder extends BaseOptimizer {
 
+  private static final Set<Mode> HANDLED_MODES =
+      Sets.newHashSet(Mode.getBu(), Mode.getb(), Mode.getIs(), Mode.getLs());
```
sherter: `ImmutableSet.of()` is preferable here.

src/main/java/minijava/ir/optimize/ConstantFolder.java
```diff
@@ -177,12 +253,20 @@ public void visit(Or node) {
 
   @Override
   public void visit(Phi node) {
-    phiStrategy.visit(node);
+    TargetValue newValue =
+        seq(node.getPreds())
+            .map(this::getValue)
+            .reduce(TargetValue.getUnknown(), ConstantFolder::supremum);
+    updateValue(node, newValue);
```
sherter, line 260: I don't quite understand why we suddenly don't need the different strategies anymore. Could you elaborate?

sgraf812: This is basically copied from the `SupremumStrategy`. I'm not sure what workaround the other strategy did, but I'm pretty confident this works.

I guess the other strategy tried to work around the fact that we also incorporated `Load`, `Store`, `Call`, `Cond`, etc. somehow. Now, these nodes are just assigned a `TargetValue.getBad()` (`defaultVisit`).

sherter: I See. Then we are good to go now 👍

PR #140, opened 2017-01-22
Always split critical edges

sgraf812:
Citing [Wikipedia](https://en.wikipedia.org/wiki/Control_flow_graph):

> A critical edge is an edge which is neither the only edge leaving its source block, nor the only edge entering its destination block. These edges must be split: a new block must be created in the middle of the edge, in order to insert computations on the edge without affecting any other edges.

The `IREmitter` will no longer emit critical edges and `JmpBlockRemover` won't remove a jump block if that would introduce a critical edge.

I also took the freedom to finally introduce a proper logging framework (`slf4j-simple`) and tied debug output to `MJ_GRAPH` (seemed like the simplest option), but we can revert that, of course.

### DIFF
--- a/src/main/java/minijava/ir/optimize/JmpBlockRemover.java
+++ b/src/main/java/minijava/ir/optimize/JmpBlockRemover.java
@@ -10,6 +10,7 @@ import firm.nodes.Jmp;
 import firm.nodes.Node;
 import java.util.List;
 import java.util.stream.Collectors;
+import minijava.ir.utils.FirmUtils;
 import minijava.ir.utils.GraphUtils;
 
 /** Removes {@link Block} nodes that contain a single {@link Jmp} node and nothing else. */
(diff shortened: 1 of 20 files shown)


src/main/java/minijava/ir/optimize/JmpBlockRemover.java
```diff
@@ -26,33 +31,47 @@ public boolean optimize(Graph graph) {
...
+    tryToIdentifyNewEdge(block)
+        .ifPresent(
+            edge -> {
+              hasChanged = true;
+              remove(block, edge.v1, edge.v2, edge.v3);
+            });
   }
 
-  private boolean isJmpBlock(Block block) {
+  private Optional<Tuple3<Node, Integer, Block>> tryToIdentifyNewEdge(Block block) {
```
sherter: I don't like this method signature very much. We can't stop you from thinking functional, can we? ;)
I'll let this pass, but next time we should really think about using Closure instead! :smile:

sgraf812: There's nothing particularly functional about tuples (Every language other than Java has them), but a separate type instead of a meaningless tuple would be great I guess.

Otherwise you have to pretty much duplicate everything in that method.

sherter: It's just the combination of Optional and Tuple with the nested generics that makes this look pretty complicated. But just leave it. We have uglier things than this one in our code... ;)

PR #142, opened 2017-01-23
Implement Load-Store optimization

sgraf812:
Eliminates consecutive Load/Load, Store/Load and Store/Store to the same addresses.


sgraf812:
Somehow `BigTensorProduct` gets optimized successfully, yet it crashes the `firm` backend:

```
Assertion failed: (bitset_is_set(available, col) && "pre-colored register must be free"), function assign, file ir/be/bechordal.c, line 360.
```

sgraf812:
This seems to break the backend on my Mac. I'm getting alignment errors again, but I'm confident I'll find the culprit.

PR #143, opened 2017-01-24
Factored out most of the stuff from Cli

sgraf812:
Cli really was an abomination. This feels much better.


src/main/java/minijava/Compiler.java
```diff
@@ -0,0 +1,309 @@
...
+import minijava.semantic.SemanticLinter;
+import minijava.token.Token;
+import minijava.util.PrettyPrinter;
+import org.apache.commons.io.FileUtils;
+import org.jetbrains.annotations.NotNull;
+import org.jooq.lambda.tuple.Tuple2;
+
+public class Compiler {
+
+  public static Lexer lex(InputStream in) {
```
sherter, line 55: This is a bit misleading, it doesn't actually `lex` the input.

sherter (approved):
Thanks for your work. There is probably much more we can do in terms of refactoring. At this point, however, I wouldn't change too much anymore before the presentation...

src/main/java/minijava/Compiler.java
sgraf812: Well, it does provide the lexed output iterator. You can't really do anything besides consuming the tokens. It's pretty much a lazily generated list of tokens.

sgraf812:
Well, this refactoring was really crucial, as there simply was too much duplication and implicit invariants. Also, I had to push the lowering of `Sel`s and `Member`s after the optimization, which was really the reason for this stuff.

src/main/java/minijava/Compiler.java
sherter: Yes, but the lazyness is not obvious. The verb `lex` suggests the opposite, which _does_ make a real difference when you think about invalid input and the time at which an error is reported. All the other methods in this class are different in this regard.

sgraf812: I think I know what you mean, but I disagree. `Iterator` really strongly suggests laziness. Also `map`, `filter` et al. are finally accepted nowadays and it's the same thing there.

PR #144, opened 2017-01-28
Alias analysis

sgraf812:
Performs a may-alias analysis on Memory dependencies.

Two pointers a and b may-alias, if in some run of the program it cannot be ruled out whether a
points to the same memory region as b.

The reverse information is pretty important: When we can prove that a and b may not alias,
they can effectively be computed independent of one another, so we don't need a transitive memory
dependency between them.

The goal of this analysis is to elimininate these unnecessary memory depenendencies, so that
 the `LoadStoreOptimizer` can optimize much more aggressively.
 
The analysis takes the following information into account:

- Flow: We calculate alias information per node.
- Types: When a points to an instance of A and b to an instance of B, we know they can't alias.
- Offsets: When we access a[0] this should obviously be independent of a[1].

I've to admit that motivation was pretty low at the end, so I made somewhat ugly hacks to have acceptable compiler performance.

I'm pretty confident that this will bring a moderate boost in benchmark performance in certain scenarios with much array/field access.


src/main/java/minijava/ir/emit/IREmitter.java
```diff
@@ -133,20 +142,29 @@ private void connectParametersToIRVariables(Method that) {
...
+      Type parameterType = storageType(p.type);
+      Node param = construction.newProj(args, parameterType.getMode(), idx);
+      NodeUtils.setLink(param, parameterType.ptr);
       construction.setVariable(idx, param);
     }
   }
 
+  private Type storageType(minijava.ast.Type type) {
+    return Types.storageType(type, classTypes);
+  }
```
sherter, line 166: I'd prefer having a `Types` object that contains the `classTypes` instead of static methods and state via parameter. That's very C. Not really related to this PR, though.

sherter:
Honestly, I really can't review this. If you _do_ want a review, you'll have to split this monster up in smaller pieces. I think some of it is already in master, but was not rebased....

sgraf812:
Yeah, probably... These polluted diffs are really unfortunate with this squash and merge style. I'll merge with master. Also, I wouldn't be mad if you didn't review this. These optimizations in itself are pretty self-contained, it's mostly the surroundings that may introduce unwanted changes (like the debatable `classTypes` stuff).

sgraf812:
Hmm... This already seems to have merged with master. Weird commit log. Regardless, I'll merge. Sorry, I hope the next bits will be more manageable.

PR #148, opened 2017-02-01
Loads don't alias with any other loads now

sgraf812:
This means that equivalent loads have identical preds and can be removed
by CSE.

This PR drastically halves the number of Loads in the hot loop of `BigTensorProduct`.

### DIFF
--- a/src/main/java/minijava/ir/optimize/AliasAnalyzer.java
+++ b/src/main/java/minijava/ir/optimize/AliasAnalyzer.java
@@ -1,15 +1,12 @@
 package minijava.ir.optimize;
 
 import static firm.bindings.binding_irnode.ir_opcode.iro_Load;
-import static firm.bindings.binding_irnode.ir_opcode.iro_Phi;
 import static firm.bindings.binding_irnode.ir_opcode.iro_Proj;
-import static firm.bindings.binding_irnode.ir_opcode.iro_Start;
 import static firm.bindings.binding_irnode.ir_opcode.iro_Store;
 import static firm.bindings.binding_irnode.ir_opcode.iro_Sync;
 import static org.jooq.lambda.Seq.seq;
 
 import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
 import firm.ArrayType;
 import firm.BackEdges;
 import firm.BackEdges.Edge;
@@ -20,9 +17,7 @@ import firm.MethodType;
 import firm.Mode;
 import firm.PointerType;
 import firm.Type;
-import firm.bindings.binding_irnode.ir_opcode;
 import firm.nodes.Address;
-import firm.nodes.Block;
 import firm.nodes.Call;
 import firm.nodes.Load;
 import firm.nodes.Member;
@@ -40,11 +35,12 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
-import minijava.ir.Dominance;
+import java.util.function.Predicate;
 import minijava.ir.emit.Types;
 import minijava.ir.utils.FirmUtils;
 import minijava.ir.utils.GraphUtils;
 import minijava.ir.utils.NodeUtils;
+import minijava.ir.utils.SideEffects;
 import org.jetbrains.annotations.NotNull;
 import org.jooq.lambda.Seq;
 import org.pcollections.HashTreePMap;
@@ -783,8 +779,6 @@ public class AliasAnalyzer extends BaseOptimizer {
 
   private class LoadStoreAliasingTransformation {
 
-    private final int MAX_NUMBER_OF_SYNC_PREDS = 10;
-
     public boolean transform() {
       // This works by following memory edges and reordering them when there is no alias.
       boolean hasChanged = false;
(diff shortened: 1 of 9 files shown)


src/main/java/minijava/ir/optimize/AliasAnalyzer.java
```diff
@@ -151,27 +151,28 @@ private void updateMemory(Node node, Memory newMemory) {
...
+  }
+
+  @Override
   public void visit(Phi node) {
-    if (node.getMode().equals(Mode.getM())) {
-      Memory memory =
-          seq(node.getPreds()).map(this::getMemory).foldLeft(Memory.empty(), Memory::mergeWith);
-      updateMemory(node, memory);
-    } else {
+    defaultVisit(node);
```
sherter: Calling `defaultVisit` first for every node should probably be enforced on an architectural level.

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
```diff
@@ -824,6 +837,7 @@ public boolean transform() {
       Node ptr = sideEffect.getPred(1);
       Set<IndirectAccess> aliasClass = getPointsTo(ptr);
       boolean isLoad = sideEffect.getOpCode() == iro_Load;
+      assert !aliasClass.isEmpty() : "Empty alias class at " + sideEffect;
```
sherter: Side note: I don't think we have assertions enabled in our integration test suite. We should probably do that.

src/main/java/minijava/ir/optimize/CommonSubexpressionElimination.java
```diff
@@ -300,12 +336,21 @@ private void binaryNode(Node node) {
   private void updateHashMapping(Node node, int hash) {
     HashedNode hashed = new HashedNode(node, hash);
     addHashedNode(hashed);
-    hashes.put(node, hashed);
+    HashedNode old = hashes.put(node, hashed);
+    if (!hashed.equals(old)) {
+      if (node.getNr() == 419) {
+        System.out.println("Changed: " + old + " new: " + hashed);
```
sherter: This debug statement shouldn't go into master.

src/main/java/minijava/ir/optimize/CommonSubexpressionElimination.java
```diff
@@ -315,10 +360,12 @@ private void addHashedNode(HashedNode hashed) {
...
-    if (Seq.of(hashedPreds).contains(null)) {
-      return Optional.empty();
+    Object[] hashedPreds = new Object[preds.length];
+    for (int i = 0; i < preds.length; ++i) {
+      hashedPreds[i] = hashes.get(preds[i]);
+      if (hashedPreds[i] == null) {
+        return Optional.empty();
+      }
     }
     return Optional.of(salt ^ Objects.hash(hashedPreds));
```
sherter: Doesn't really belong to this PR, but `salt` should be added as an argument to the varargs Objects.hash` method.

sherter (changes requested):

sherter:
Removing the debug statement will be enough to get my approval; the others are just suggestions.

src/main/java/minijava/ir/optimize/CommonSubexpressionElimination.java
sgraf812: You're right, I thought about it when I hit that line. But that would mean we'd have to wrap the entire call in a new array, something I didn't really want to do. We should be good anyways because collisions are resolved by the `ExtensionalEqualityComparator`.

Edit: Whatever, done.

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
sgraf812: I though about just calling `super.visit(node)`, which does the same, but I'd figured it would be more explicit what's going on this way.

sgraf812:
I'm not sure which debug statements you mean? I screwed that up on my last PR, I admit, but I took extra care to delete everything this time. At least I couldn't find anything by grepping for `System.out`.

Thanks for the review(s), btw. :)

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
sgraf812: I *think* they are enabled in the `run` script. At least it works when going through `mjtest`.

src/main/java/minijava/ir/optimize/CommonSubexpressionElimination.java
sgraf812: Sorry, now I know what you meant. This was based on an old commit :)

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
sherter: How about a putting your visitor into an `Pre/PostProcessingVisitor`? You are a fan of composition, aren't you? :wink:

sgraf812: Well, it's not called for *every* node (e.g. `Call`). Also I think that this is already DRY enough.

sgraf812: I could probably pull out the logic of `defaultVisit` into `mergeMemoryFromPreds` and call that instead, sounds the like the most reasonable thing to do.

sherter:
Ah, you removed it again in a later commit, it's all right then. I usually use commits as a way to logically group things and prefer to rewrite history instead of appending another commit later on, but I guess that's personal taste.

sherter (approved):

sgraf812:
Yeah, it's probably bad practice, but it's only a working branch. Also I don't think rewriting history would've helped, as your last pull was probably before the commit that fixed it. Rewriting history is also fishy when there are multiple repos that already fetched the refs (as it would be in this case), because one would've to delete the branch/`reset --hard origin/branch`. 

This could've been prevented if I'd just waited a little longer with the PR, though...

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
sherter: Right, @parttimenerd enabled this here: https://github.com/mj3-16/minijavac/blame/ffe577850cc820dc72c32f140124f352938e1967/build.gradle#L46
Was probably in one of the PRs I didn't really review ;)
Enabling it always by default might not be the best idea, though. I'm not sure how big the runtime cost is, but later in the competition this might make the difference ;)

sgraf812: Judging from where I had to fight bottlenecks so far, assertions never were a concern.

I think we should disable them nonetheless for the benchmarks, because it would be really annoying if our compiler broke because of one of them. For example in this case it might be possible to continue, taking the odds of incorrectly generated code (which only really broke for the `big/math.java` test case).

parttimenerd: We could check a environment variable (e.g. `MJ_ASSERTIONS`) for enabling assertions in the `run` script and set the environment variable by default in all testing scenarios.

PR #149, opened 2017-02-03
Loop invariant code motion

sgraf812:
Done through loop inversion: Copying the loop header and inserting the invariant code there, while the back edge will still point to the old header, which won't repeat the computation. That effectively transforms a while loop into a do/while loop, enabling the invariant code to be moved in a block outside the loop that is post-dominated by its old definition block. Which in turn guarantees that we only ever compute the value if it would be before the transform.

This is all based on 'Compiler Design: Analysis and Transformation' and Apple's 'Modern Compiler Implementation in ML'.


sgraf812:
In case anyone is wondering, this isn't actually 35 commits, more like 15. This was branched off master before the last PR, so the log intermingles commits because of the Squash & merge workflow. These will not be seen on master when I squash them.

PR #150, opened 2017-02-03
Added parameters for log level and rudimentary optimization level

sgraf812:
`./run --optimize 3 -v 3 file.java`.

Doesn't really differentiate between optimization levels 1-3 yet, because the pending PR will change that anyway.

Also I don't feel motivated to refactor our `Cli` class right now, but it's definitely using `jCommander` in a wrong way.

### DIFF
--- a/src/main/java/minijava/Cli.java
+++ b/src/main/java/minijava/Cli.java
@@ -266,6 +266,10 @@ public class Cli {
     /** Returns true if the parameter values represent a valid set */
     boolean valid() {
       return !invalid
+          && 0 <= optimizationLevel
+          && optimizationLevel <= 3
+          && 0 <= verbosity
+          && verbosity < Level.values().length
           && (help
               || ((Booleans.countTrue(
                           echo, lextest, parsetest, printAst, check, compileFirm, runFirm, printAsm)
--- a/src/main/java/minijava/Cli.java
+++ b/src/main/java/minijava/Cli.java
@@ -26,6 +26,7 @@ import minijava.Compiler.Backend;
 import minijava.ast.Program;
 import minijava.lexer.Lexer;
 import minijava.token.Token;
+import org.slf4j.event.Level;
 import org.slf4j.impl.SimpleLogger;
 
 public class Cli {
@@ -34,7 +35,7 @@ public class Cli {
       Joiner.on(System.lineSeparator())
           .join(
               new String[] {
-                "Usage: minijavac [--echo|--lextest|--parsetest|--check|--compile-firm|--print-asm] [--help] file",
+                "Usage: minijavac [--echo|--lextest|--parsetest|--check|--compile-firm|--print-asm] [--optimize] [--help] [--verbosity] file",
                 "",
                 "  --echo          write file's content to stdout",
                 "  --lextest       run lexical analysis on file's content and print tokens to stdout",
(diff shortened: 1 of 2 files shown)


sherter:
> but it's definitely using jCommander in a wrong way.

Could you be more specific?

src/main/java/minijava/Cli.java
```diff
@@ -44,6 +45,8 @@
                 "  --compile-firm  compile the given file with the libfirm amd64 backend",
                 "  --run-firm      compile and run the given file with libfirm amd64 backend",
                 "  --print-asm     compile the given file and output the generated assembly",
+                "  --optimize|-O   Optimization level of produced IR. 0-3",
+                "  --verbosity|-v  Crank this up for more debug output",
```
sherter: Please validate (in `Parameters.validate()`) that the integers are inside this range.

src/main/java/minijava/Cli.java
```diff
@@ -267,10 +277,9 @@ static Parameters parse(String... args) {
       Parameters params = new Parameters();
       try {
         new JCommander(params, args);
-        if (params.mainParameters.size() == 1) {
-          params.file = params.mainParameters.get(0);
-        }
+        params.file = params.mainParameters.get(params.mainParameters.size() - 1);
```
sherter: This change makes it possible to pass multiple files to the CLI without failing (in `Parameters.validate()`), please revert this.

sherter (changes requested):

src/main/java/minijava/Cli.java
sherter: Since you've decided to use some other workaround to let the test pass, I'll try to make this more clear:
This line is plain _wrong_. It will throw an `IndexOutOfBoundsException` if no "main argument" is was given.

sgraf812: Ah, of course.

sherter: Really, just go back to the original version where `file` is only set if it is unambiguous and remove [line 278](https://github.com/mj3-16/minijavac/pull/150/files#diff-621504ca28f913dedbeacbc5dfbcea20R278) (which I really don't understand, how is `echo` special?)

sgraf812: `echo` is special in that a test broke.

Really, I'm not motivated to dig into this code any longer, I just want to make it work good enough. The CLI works as is (probably?!) and any more changes should really refactor this more into a style that `jCommander` expects.

The reason why I did this, is so that you can optionally specify `--optimize` and `--verbosity`. There's no other way around this not involving major code changes, in a part of the compiler irrelevant to the actual task.

I don't like what I did there (starting from something battle-tested, but spaghettish and make it even fuglier), but boasting with our CLI code is not something I plan to when this is finished.

sherter:
I hope you don't mind that I streamlined your changes a bit.
I understand why you don't care too much, but integrating something that I know is bad just doesn't feel right ;)

sgraf812:
No problem, thank you. But I still don't get why we get away with that `params.mainParameters.size() == 1`. I was under the impression that `mainParameters` contained all command line arguments, which is why I made that change to support `--compile-firm --optimize 2 --verbosity 1 file.java`, which would have more than one `mainParameter`, I guess?!

sherter:
`mainParameters` doesn't contain _all_ command line arguments, it only contains those which are not options (in our case only `file`). The `2` and `1` in your example is directly assigned to `optimizationLevel` and `verbosity`, respectively.

sgraf812:
OK, so I got it all wrong to begin with. Thanks for clearing that up.

sherter (approved):

PR #153, opened 2017-02-06
Run intraprocedural optimizations only if necessary

sherter:
Previously, we re-ran the whole intra-procedural optimization
framework on all graphs after having done inter-procedural
optimizations.
Now, we re-run the intra-procedural optimizations only on those graphs
which were actually changed by inter-procedural optimizations.


sgraf812:
You're free to merge this. I think that the changes are so minimal that one approval should be enough.

PR #154, opened 2017-02-07
Reverted LoadStoreOptimizer.java

sgraf812:
… to the state of 5300473a88685cd2367a23fb2f1e8e865295476a, which should be correct in absence of Loop Invariant Code Motion (and also in presence, but LICM probably screwed something up).

Edit: Well, it wasn't. Now we're taking into account data dependencies on `store.getValue()` and thus should be fine, while still being able to perform the important transformations.

### DIFF
--- a/src/main/java/minijava/ir/utils/GraphUtils.java
+++ b/src/main/java/minijava/ir/utils/GraphUtils.java
@@ -61,19 +61,17 @@ public class GraphUtils {
   }
 
   public static boolean areConnected(Node source, Node target) {
+    final boolean[] connected = {false};
     Consumer<Node> visitor =
         n -> {
           if (n.equals(target)) {
-            throw new EarlyExitException();
+            connected[0] = true;
           }
         };
 
-    try {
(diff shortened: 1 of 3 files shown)


src/main/java/minijava/ir/utils/GraphUtils.java
```diff
@@ -61,17 +61,19 @@
...
-    walkFromNodeDepthFirst(source, visitor, n -> {});
-
-    return connected[0];
+    try {
+      walkFromNodeDepthFirst(source, visitor, n -> {});
+      return false;
+    } catch (EarlyExitException e) {
+      return true;
+    }
   }
```
sherter: I don't particularly like misusing exceptions for manipulating control flow. I can see why you think this is faster, but is it? Did you measure it?

sgraf812: Yeah, I know. I really hesitated with this and no, I haven't measured. I should revert it.

PR #157, opened 2017-02-12
Various improvements to reliable compiler performance

sgraf812:
- Identify loop breakers in the call graph which are never inlined
- Actually differentiate optimization levels (0 = only perform slight optimizations just before calling the backend, 1 = typical optimizations, 2 = also perform alias analysis)
- Recompute dominance only when necessary through explicit invalidation (huge boost)
- Use `instanceof` instead of `getOpcode()` in checks (moderate)
- Don't activate/deactivate out edges in hot spots (moderate)

This means much better worst case compiler performance, so that we should rarely time out.

I also tried to run the alias analysis after inlining only, but that didn't improve compiler performance as much as it slowed down produced code.

`-v 2` will now print performance statistics of each run optimization.

### DIFF
--- a/src/main/java/minijava/ir/optimize/AliasAnalyzer.java
+++ b/src/main/java/minijava/ir/optimize/AliasAnalyzer.java
@@ -1,9 +1,5 @@
 package minijava.ir.optimize;
 
-import static firm.bindings.binding_irnode.ir_opcode.iro_Load;
-import static firm.bindings.binding_irnode.ir_opcode.iro_Proj;
-import static firm.bindings.binding_irnode.ir_opcode.iro_Store;
-import static firm.bindings.binding_irnode.ir_opcode.iro_Sync;
 import static org.jooq.lambda.Seq.seq;
 
 import com.google.common.collect.Lists;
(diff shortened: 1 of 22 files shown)


src/main/java/minijava/ir/optimize/AliasAnalyzer.java
```diff
@@ -107,19 +93,16 @@ public boolean optimize(Graph graph) {
       return false;
     }
     LoadStoreAliasingTransformation transformation = new LoadStoreAliasingTransformation();
-    return FirmUtils.withBackEdges(graph, transformation::transform);
+    Boolean aBoolean = FirmUtils.withBackEdges(graph, transformation::transform);
+    return aBoolean;
```
sherter, line 97: Seems unnecessary (but no need to change it)

sherter (approved):
Only had a quick look over it, but seems to be okay

src/main/java/minijava/ir/optimize/AliasAnalyzer.java
sgraf812: Ah, this was a leftover from debugging.

